PR #1, opened 2020-04-15
Add support for black config

Mystic-Mirage:
Hi!

Now darker can use the black config.

It's ugly because it uses a dirty hack with click module. This was done for reusing purposes of the black code.


akaihola:

akaihola:

Mystic-Mirage:
Changed, according requests

PR #4, opened 2020-06-13
Add a --diff option to match black behavior.

Carreau:
This allows for example to use darker as a formatter for vscode with
minimal changes as vscode sends the --quiet and --diff flags to the
formatter.

    "python.formatting.provider": "black",
    "python.formatting.blackPath": "/path/to/darker"


Carreau:
A bit ugly of an implementation, mainly to know if you would be ok with it. 

I was also happy to find this project as I was really unhappy with inserting/removing `format:on/off` in darken, and ended up with the same diff idea as you and found you implemented it.

Carreau:
Also just deprecated darken and point to darken now.
(https://pypi.org/project/darken/)

akaihola:
Hey @Carreau this is a great contribution! I've started to review – please expect a couple of suggestions.

src/darker/__main__.py
```diff
@@ -107,7 +107,24 @@ def format_edited_parts(srcs: Iterable[Path], isort: bool) -> None:
                 # 10. A re-formatted Python file which produces an identical AST was
                 #     created successfully - write an updated file
                 logger.info("Writing %s bytes into %s", len(result_str), src)
-                src.write_text(result_str)
+                if print_diff:
+                    from difflib import unified_diff
+
+                    difflines = list(
+                        unified_diff(
+                            open(src).read().splitlines(),
```
akaihola, line 115: I think the `edited` variable already contains the lines of the edited file before reformatting.
```suggestion
                            edited,
```

src/darker/__main__.py
```diff
@@ -107,7 +107,24 @@ def format_edited_parts(srcs: Iterable[Path], isort: bool) -> None:
...
+                            result_str.splitlines(),
```
akaihola, line 116: Also, `chosen_lines` already contains the lines of the resulting file after picking the correct reformatted lines.

I'm not sure what the exact behavior there is regarding the final linefeed in the file, could you try and check?

```suggestion
                            chosen_lines,
```

src/darker/__main__.py
```diff
@@ -107,7 +107,24 @@ def format_edited_parts(srcs: Iterable[Path], isort: bool) -> None:
                 # 10. A re-formatted Python file which produces an identical AST was
                 #     created successfully - write an updated file
                 logger.info("Writing %s bytes into %s", len(result_str), src)
-                src.write_text(result_str)
+                if print_diff:
+                    from difflib import unified_diff
```
akaihola, line 111: Is there a specific reason to do a local import?

src/darker/__main__.py
```diff
@@ -107,7 +107,24 @@ def format_edited_parts(srcs: Iterable[Path], isort: bool) -> None:
...
+                    difflines = list(
+                        unified_diff(
+                            open(src).read().splitlines(),
+                            result_str.splitlines(),
+                            src.as_posix(),
+                            src.as_posix(),
+                        )
+                    )
+                    if len(difflines) > 2:
+                        h1, h2, *rest = list(difflines)
```
akaihola, line 122: No need for `list()`, `difflines` is already a list (see line 113).

```suggestion
                        h1, h2, *rest = difflines
```

src/darker/command_line.py
```diff
@@ -27,6 +27,9 @@ def parse_command_line(argv: List[str]) -> Namespace:
     isort_help = ["Also sort imports using the `isort` package"]
     if not isort:
         isort_help.append(f". {ISORT_INSTRUCTION} to enable usage of this option.")
+    parser.add_argument(
+        "--diff", action="store_true", help="print diff instead of modifying the file",
```
akaihola, line 31: All the other command help texts start with a title case word.
```suggestion
        "--diff", action="store_true", help="Print diff instead of modifying the file",
```

akaihola:

akaihola:

Carreau:
> There you go, just a few minor notes to keep the code style consistent and skip some unnecessary operations.
> 
> Do you think you could add a simple test case for the `--diff` option?

Yes, of course, I mainly wanted to test the waters. I'll cleanup, and I'll also see if I can make it work with isort, which I'm not sure it does now.

akaihola:
A section for VSCode in `README.rst` / "Editor integration" would also be great after this feature is done!

src/darker/command_line.py
akaihola: Actually, let's make the text identical to that in `black --help`

PR #15, opened 2020-07-04
Compatibility with isort 5.x

akaihola:
- [ ] ~~wait for #13 to be merged, this one is based on that~~
- [ ] ~~wait for [pytest-isort](https://github.com/moccu/pytest-isort/) to [fix compatibility with isort 5.x](https://github.com/moccu/pytest-isort/issues/23)~~
- [ ] ~~re-enable `pytest-isort` in `setup.cfg` and `pytest.ini`~~

Let's use #16 instead and discard this PR. But I'll keep this open for now so we can easily see the diff in case there's something useful there.


akaihola:
I based this on top of `isort_config` so the diff only shows changes on top of that.

Mystic-Mirage:
@akaihola I don't want to #13 to be merged before isort>5 because new isort allows writing less of code.

PR #16, opened 2020-07-04
Add support of isort>5

Mystic-Mirage:
- [ ] re-enable pytest-isort after moccu/pytest-isort/pull/24

### DIFF
--- a/src/darker/import_sorting.py
+++ b/src/darker/import_sorting.py
@@ -3,7 +3,7 @@ import logging
 try:
     import isort
 except ImportError:
-    isort = False
+    isort = None
 
 logger = logging.getLogger(__name__)
 
--- a/src/darker/import_sorting.py
+++ b/src/darker/import_sorting.py
@@ -1,24 +1,16 @@
 import logging
-from pathlib import Path
-from typing import List, cast
 
 try:
-    from isort import SortImports
+    import isort
 except ImportError:
-    SortImports = None
+    isort = False
 
 logger = logging.getLogger(__name__)
 
 
 def apply_isort(content: str) -> str:
-    logger.debug(
-        "SortImports(file_contents=..., check=True, multi_line_output=3, "
-        "include_trailing_comma=True, force_grid_wrap=0, use_parentheses=True, "
-        "line_length=88, quiet=True)"
-    )
-    result = SortImports(
-        file_contents=content,
-        check=True,
+    isort.check_code(code=content)
+    isort_config_kwargs = dict(
         multi_line_output=3,
         include_trailing_comma=True,
         force_grid_wrap=0,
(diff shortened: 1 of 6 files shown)


akaihola:
Oops, I have a competing PR in #15 :smile:

Mystic-Mirage:
> Oops, I have a competing PR in #15

I'd prefer to add support of isort>5 first and then reimplement #13

Mystic-Mirage:
@akaihola New `isort` makes config processing much easier for us.

akaihola:
I think I managed to reduce unit test hassle a bit by using `isort.api.sort_code_string()` instead of `isort.check_code()` and `isort.code()` in #15.

src/darker/import_sorting.py
```diff
@@ -1,29 +1,27 @@
...
 def apply_isort(content: str) -> str:
-    logger.debug(
-        "SortImports(file_contents=..., check=True, multi_line_output=3, "
-        "include_trailing_comma=True, force_grid_wrap=0, use_parentheses=True, "
-        "line_length=88, quiet=True)"
-    )
-    result = SortImports(
-        file_contents=content,
-        check=True,
+    isort.check_code(code=content)
```
akaihola, line 12: Why is this call needed?

Mystic-Mirage: This replaces `check=True` and writes error messages to `stdout`. See changes in `test_main`.

akaihola: Do we need the error messages?

And if we do, is there a cleaner way to capture them than `stdout`?

akaihola:
We'll need this in `setup.cfg`:
```ini
[options.extras_require]
isort =
    isort>=5.0.1
```

(There seems to be a bug in `isort` 5.0.0)

Mystic-Mirage:
> I had to solve a couple of unit test challenges in #15 – maybe looking at those will help you then with #13?

All tests pass after commenting `--isort`

src/darker/import_sorting.py
Mystic-Mirage: > Do we need the error messages?

I don't know... Some tests depend on error messages. Probably, we could raise an exception if `check_code` is `False`

> And if we do, is there a cleaner way to capture them than `stdout`?

Hm... `isort` has hardcoded prints to write to `stdout`

akaihola: Well, I had to include those error messages in the test assertions, because in `isort<5` there was no way to get rid of the messages.

Now that we can run `isort` without polluting `stdout` with error messages, I think we should do that for simplicity's sake, and fix the tests as well – unless we get some benefit from having those error messages.

Mystic-Mirage: I thought `check=False` in `isort<5` was for this?
I vote to get rid of error messages in another PR in order to keep this PR clear.

akaihola: > I vote to get rid of error messages in another PR in order to keep this PR clear.

I created another PR: #17

PR #18, opened 2020-07-06
Allow to configure `isort` through `pyproject.toml`

Mystic-Mirage:
Attempt # 2. Decided to rewrite #13 from scratch after #16 due to a lot of conflicts and redundant code

@akaihola your comments from #13 were taken to account:

- TypedDict for BlackArgs / IsortArgs
- debug message is readable enough

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,3 +11,4 @@ include_trailing_comma = true
 force_grid_wrap = 0
 use_parentheses = true
 line_length = 88
+known_third_party = ["pytest"]
--- a/src/darker/tests/test_import_sorting.py
+++ b/src/darker/tests/test_import_sorting.py
@@ -11,7 +11,7 @@ ISORTED_SOURCE = "import os\nimport sys\n"
 
 
 def test_apply_isort():
-    result = apply_isort(ORIGINAL_SOURCE)
+    result = apply_isort(ORIGINAL_SOURCE, Path("test1.py"))
 
     assert result == ISORTED_SOURCE
 
@@ -58,8 +58,7 @@ def test_isort_config(monkeypatch, tmpdir, line_length, settings_file, expect):
     )
 
     content = "from module import ab, cd, ef, gh, ij, kl, mn, op, qr, st, uv, wx, yz"
-    src = Path(tmpdir / "test1.py") if not settings_file else None
     config = str(tmpdir / settings_file) if settings_file else None
 
-    actual = apply_isort(content, src, config)
+    actual = apply_isort(content, Path("test1.py"), config)
     assert actual == expect
--- a/src/darker/tests/test_import_sorting.py
+++ b/src/darker/tests/test_import_sorting.py
@@ -1,3 +1,8 @@
+from pathlib import Path
+from textwrap import dedent
+
+import pytest
+
 from darker.import_sorting import apply_isort
 
 ORIGINAL_SOURCE = "import sys\nimport os\n"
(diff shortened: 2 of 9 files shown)


pyproject.toml
```diff
@@ -11,3 +11,4 @@ include_trailing_comma = true
 force_grid_wrap = 0
 use_parentheses = true
 line_length = 88
+known_third_party = ["pytest"]
```
Mystic-Mirage, line 14: Without this `isort` puts `pytest` after `darker` in my environment

src/darker/tests/test_import_sorting.py
```diff
@@ -8,3 +13,28 @@ def test_apply_isort():
...
+        )
+    )
+
+    content = "from module import ab, cd, ef, gh, ij, kl, mn, op, qr, st, uv, wx, yz"
+    src = Path(tmpdir / "test1.py") if not settings_file else None
+    config = str(tmpdir / settings_file) if settings_file else None
+
+    actual = apply_isort(content, src, config)
+    expected = apply_isort(content, line_length=line_length)
+    assert actual == expected
```
akaihola, line 40: Ok, I see now where the need for `src: Optional[Path]` came from.

My suggestion would be to have explicit expected results instead of relying on calling the tested function twice with different arguments to verify its own output. If a bug is introduced in that function, it could very well be that both of those use cases are affected, and the test reveals nothing!

This also could make it more obvious to the reader of the code what actually is happening.

```suggestion
@pytest.mark.parametrize(
    "line_length, settings_file, expect",
    [
        (
            50,
            None,
            "from module import (ab, cd, ef, gh, ij, kl, mn,\n"
            "                    op, qr, st, uv, wx, yz)\n",
        ),
        (
            50,
            "pyproject.toml",
            "from module import (ab, cd, ef, gh, ij, kl, mn,\n"
            "                    op, qr, st, uv, wx, yz)\n",
        ),
        (
            60,
            None,
            "from module import (ab, cd, ef, gh, ij, kl, mn, op, qr, st,\n"
            "                    uv, wx, yz)\n",
        ),
        (
            60,
            "pyproject.toml",
            "from module import (ab, cd, ef, gh, ij, kl, mn, op, qr, st,\n"
            "                    uv, wx, yz)\n",
        ),
    ],
)
def test_isort_config(monkeypatch, tmpdir, line_length, settings_file, expect):
    find_project_root.cache_clear()
    monkeypatch.chdir(tmpdir)
    (tmpdir / 'pyproject.toml').write(
        dedent(
            f"""\
            [tool.isort]
            line_length = {line_length}
            """
        )
    )

    content = "from module import ab, cd, ef, gh, ij, kl, mn, op, qr, st, uv, wx, yz"
    src = Path(tmpdir / "test1.py") if not settings_file else None
    config = str(tmpdir / settings_file) if settings_file else None

    actual = apply_isort(content, src, config)
    assert actual == expect
```

src/darker/tests/test_import_sorting.py
```diff
@@ -1,3 +1,8 @@
+from pathlib import Path
+from textwrap import dedent
+
+import pytest
```
akaihola, line 4: I'd prefer all imports at the top for clarity unless a local import is required to avoid import loops.

```suggestion
import pytest
from black import find_project_root
```

src/darker/tests/test_main.py
```diff
@@ -39,15 +44,14 @@ def test_isort_option_with_isort(run_isort):
     assert "Please run" not in run_isort.caplog.text
 
 
-def test_isort_option_with_isort_calls_sortimports(run_isort):
+@pytest.mark.parametrize(
+    "run_isort,isort_args",
```
akaihola, line 48: I've preferred to always delimit the `parametrize` argument list with spaces so it nicely matches the function signature visually.

Cool trick with `request.param`! I had forgotten about the ability to pass parameters to fixtures. Thanks!

```suggestion
    "run_isort, isort_args",
```

src/darker/tests/test_main.py
```diff
@@ -22,14 +22,19 @@ def test_isort_option_without_isort(tmpdir, without_isort, caplog):
 
 
 @pytest.fixture
-def run_isort(git_repo, monkeypatch, caplog):
+def run_isort(git_repo, monkeypatch, caplog, request):
+    from black import find_project_root
```
akaihola, line 26: Here, too, maybe move import to top if there's no reason to have a local import?

PR #28, opened 2020-07-14
Linenum diffing refactor

akaihola:
Refactor the algorithm for finding the list of line numbers of modified lines. Moving it into a new `EditedLinenumsDiffer` class in `darker.git`. This is also in preparation for adding possible support for linters (like Mypy and Pylint).


src/darker/__main__.py
```diff
@@ -50,54 +46,44 @@ def format_edited_parts(
...
-        edited_srcs = {
-            src: apply_isort(edited_content, src, config, line_length)
-            for src, edited_content in worktree_srcs.items()
-        }
-    else:
-        edited_srcs = worktree_srcs
-
-    for src_relative, edited_content in edited_srcs.items():
-        max_context_lines = len(edited_content)
+    edited_linenums_differ = EditedLinenumsDiffer(git_root)
```
samoylovfp, line 49: I'd inline this or move this line to the only usage of this object (before line 70)

PR #36, opened 2020-08-04
`--check` option, mirroring `black --check`

akaihola:
`--check` skips writing changes to files, and returns a return value of 1 from the `darker` process in case any modified lines in any of the given files are not already properly formatted.

The return value will allow using `darker` with more different tools. It for example opens the possibility for a `pytest-darker` plugin.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -6,8 +6,12 @@ These features will be included in the next release:
 Added
 -----
 
+- ``--check`` returns 1 from the process but leaves files untouched if any file would
+  require reformatting
+
 Fixed
 -----
+
 
 1.0.0_ - 2020-07-15
 ===================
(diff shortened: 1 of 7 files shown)


akaihola:
I rebased on `master` and resolved conflicts.

akaihola:
@samoylovfp, both `--diff` and `--check` do imply that changes aren't written into files. The reason for this is simply that it's how Black works. This ensures that Darker can more easily act as a direct replacement for Black when integrating with other tools.

akaihola:
> I would refactor the code so there is a function
> that returns the proposed changes
> and another function that actually applies those

I did that, could you make a another quick check on the code? Quite some test changes were required as a result.

samoylovfp:
@akaihola I think it looks excellent!

PR #40, opened 2020-08-04
Link other interesting projects in the README

akaihola:


Carreau:
I'm also working on https://github.com/Carreau/velin, which reformat docstrings that follow the numpydoc convention, (and also can run black in docstrings tha follow numpydoc convention with `>>>` and `...` prompts.

PR #43, opened 2020-08-07
WIP: Add -r / --revision argument for specifying which Git revision to compare to

### LINKED ISSUE #42: Option for specifying Git revision to compare against
Darker now always compares the working tree against Git `HEAD` to find out which lines were changed. It would be useful to allow comparisons against other revisions as well. One use case is to compare a feature branch to the branching point so formatting and linting fixes can be done before merging.

This could be implemented with a `-r` / `--revision` command line argument, defaulting to `HEAD`, and being used in `darker.git.git_get_unmodified_content()`. That function would include the revision when calling `git show <revision>:/<path.py>`.

akaihola:
Fixes #42. Should be reviewed and merged only after #48 has been merged.

TODO: unit tests
- [x] two commits in a repo, compare to first
- [x] test case for when current files didn't exist at the given revision
- [x] test case for `git_get_unmodified_content()` with a non-`HEAD` revision
- [x] add cases to `test_revision()` with unmodified content compared to the given revision

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -113,7 +113,9 @@ The following `command line arguments`_ can also be used to modify the defaults:
 .. code-block:: shell
 
      -r REVISION, --revision REVISION
-                           Git revision against which to compare the working tree
+                           Git revision against which to compare the working
+                           tree. Tags, branch names, commit hashes, and other
+                           expressions like HEAD~5 work here.
 
      --diff                Don't write the files back, just output a diff for
                            each file on stdout. Highlight syntax on screen if
--- a/README.rst
+++ b/README.rst
@@ -112,6 +112,9 @@ The following `command line arguments`_ can also be used to modify the defaults:
 
 .. code-block:: shell
 
+     -r REVISION, --revision REVISION
+                           Git revision against which to compare the working tree
+
      --diff                Don't write the files back, just output a diff for
                            each file on stdout. Highlight syntax on screen if
                            the `pygments` package is available.
@@ -137,9 +140,11 @@ The following `command line arguments`_ can also be used to modify the defaults:
 
 *New in version 1.0.0:* isort_ is configured with ``-c`` and ``-l``, too.
 
-*New in version 1.1.0:* The ``--check`` command line option.
+*New in version 1.1.0:* The ``-r`` / ``--revision`` command line option.
 
 *New in version 1.1.0:* The ``--diff`` command line option.
+
+*New in version 1.1.0:* The ``--check`` command line option.
 
 *New in version 1.1.0:* The ``--no-skip-string-normalization`` command line option.
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,6 +26,7 @@ packages = find:
 install_requires =
     black
     typing-extensions ; python_version < "3.8"
+    dataclasses ; python_version < "3.7"
 python_requires = >=3.6
 
 [options.packages.find]
--- a/src/darker/__main__.py
+++ b/src/darker/__main__.py
@@ -19,7 +19,7 @@ logger = logging.getLogger(__name__)
 
 
 def format_edited_parts(
-    srcs: Iterable[Path], enable_isort: bool, black_args: BlackArgs
+    srcs: Iterable[Path], revision: str, enable_isort: bool, black_args: BlackArgs
 ) -> Generator[Tuple[Path, str, str, List[str]], None, None]:
     """Black (and optional isort) formatting for chunks with edits since the last commit
 
@@ -39,6 +39,7 @@ def format_edited_parts(
     10. write the reformatted source back to the original file
 
     :param srcs: Directories and files to re-format
+    :param revision: The Git revision against which to compare the working tree
     :param enable_isort: ``True`` to also run ``isort`` first on each changed file
     :param black_args: Command-line arguments to send to ``black.FileMode``
     :return: A generator which yields details about changes for each file which should
@@ -46,8 +47,8 @@ def format_edited_parts(
 
     """
     git_root = get_common_root(srcs)
-    changed_files = git_get_modified_files(srcs, git_root)
-    edited_linenums_differ = EditedLinenumsDiffer(git_root)
+    changed_files = git_get_modified_files(srcs, revision, git_root)
+    edited_linenums_differ = EditedLinenumsDiffer(git_root, revision)
 
     for path_in_repo in changed_files:
         src = git_root / path_in_repo
@@ -68,7 +69,7 @@ def format_edited_parts(
         for context_lines in range(max_context_lines + 1):
             # 2. diff HEAD and worktree for the file
             # 3. extract line numbers in each edited to-file for changed lines
-            edited_linenums = edited_linenums_differ.head_vs_lines(
+            edited_linenums = edited_linenums_differ.revision_vs_lines(
                 path_in_repo, edited_lines, context_lines
             )
             if (
@@ -199,7 +200,7 @@ def main(argv: List[str] = None) -> int:
     # We need both forms when showing diffs or modifying files.
     # Pass them both on to avoid back-and-forth conversion.
     for path, old_content, new_content, new_lines in format_edited_parts(
-        paths, args.isort, black_args
+        paths, args.revision, args.isort, black_args
     ):
         some_files_changed = True
         if args.diff:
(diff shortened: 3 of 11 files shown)


Carreau:
Ah well, I see we started working on the same thing at the same time twice !

Carreau:
I was also thinking in a config file (pyproject.toml at the root of the repo?) having the ability to give a revision to be used by default. But +1.

akaihola:
>Ah well, I see we started working on the same thing at the same time twice !

:smile: yes... the solutions are pretty similar.

Do you know if there's a difference in any situation between `git show :./file.py` and `git show HEAD:./file.py`?
I did notice that in one of the tests I did only `git add` without `git commit`, and that test broke when `:` changed to `HEAD:`.

I'm also considering different names for the option. Git documentation shows `<commit>` in command line examples, explains the syntax in `"SPECIFYING REVISIONS"`, and I also see "commit-ish" and "committish" used.

From the glossary:

>**[commit](https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefcommitacommit)**
>As a noun: A single point in the Git history; the entire history of a project is represented as a set of interrelated commits. The word "commit" is often used by Git in the same places other revision control systems use the words "revision" or "version". Also used as a short hand for commit object.
>
>**[commit-ish](https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefcommit-ishacommit-ishalsocommittish)** (also committish)
>A commit object or an object that can be recursively dereferenced to a commit object. The following are all commit-ishes: a commit object, a tag object that points to a commit object, a tag object that points to a tag object that points to a commit object, etc.
>
>**[revision](https://git-scm.com/docs/gitglossary#Documentation/gitglossary.txt-aiddefrevisionarevision)**
Synonym for commit (the noun).

So I see your point about "commit-ish" – for example a tag object is not a commit or a revision.

But I also think `--commitish` [sic] is a clumsy option name and probably an obscure term for the average user.  That's why I ended up using `--revision` instead – also if you look at the [gitrevisions](https://git-scm.com/docs/gitrevisions) documentation, that [does use `<rev>`](https://git-scm.com/docs/gitrevisions#Documentation/gitrevisions.txt-emltrevgtltpathgtemegemHEADREADMEememmasterREADMEem) for the part of the `git show` expression we're using:

    <rev>:<path>, e.g. HEAD:README, master:./README

These are the reasons I came up with `--revision`. What do you think?

akaihola:
Summary of the main differences between mine vs. your solution:
- `revision` vs. `commitish`
- rename `head_vs_lines()` to `revision_vs_lines()` vs. keep the name
- default value of `"HEAD"` vs. `""`
- handle missing files in the given revision (in my version)

The rest of the differences are mostly a matter of taste. @Carreau could you review my MR and point out any flaws and design decisions you dislike? Also if you accept my invite on GitHub to become a collaborator on this repository, I can assign the MR for you to review.

akaihola:
@Carreau, I found a flaw in that untracked files were completely skipped. That's now fixed in #48, and this branch was re-built on top of that. Could you also review #48?

Carreau:
> Do you know if there's a difference in any situation between `git show :./file.py` and `git show HEAD:./file.py`?

Yeah, my main concern would be if users have staged only some ; or partial version of a file I think it _may_ differ; but I'm not too sure. I think that's fine enough and can be refined in further iteration if there are complaint.

I actually like `-r/--revisions`, I think it will be understandable enough, and also acomodate for potentially other VCS later (though I'm doubtfull), it's neet that it seem to work with things like `@{yesterday}`, or `master...`.

README.rst
```diff
@@ -112,6 +112,8 @@ The following `command line arguments`_ can also be used to modify the defaults:
 
 .. code-block:: shell
 
+     -r REVISION, --revision REVISION
+                           Git revision against which to compare the working tree
```
Carreau, line 116: I would add "should work with `HEAD~1`, `master..`, and other git expressions", or other similar phrasing, for people to understand it's not only hash based.

src/darker/__main__.py
```diff
@@ -39,17 +39,18 @@ def format_edited_parts(
...
 
     """
     git_root = get_common_root(srcs)
-    changed_files = git_diff_name_only(srcs, git_root)
-    edited_linenums_differ = EditedLinenumsDiffer(git_root)
+    changed_files = git_get_modified_files(srcs, revision, git_root)
+    edited_linenums_differ = EditedLinenumsDiffer(git_root, revision)
 
-    for path_in_repo in changed_files:
+    for path_in_repo in sorted(changed_files):
```
Carreau, line 53: 👍

src/darker/git.py
```diff
@@ -1,64 +1,116 @@
...
+    cmd = ["git", "show", f"{revision}:./{path}"]
     logger.debug("[%s]$ %s", cwd, " ".join(cmd))
-    return check_output(cmd, cwd=str(cwd), encoding='utf-8').splitlines()
+    try:
+        return check_output(cmd, cwd=str(cwd), encoding='utf-8').splitlines()
+    except CalledProcessError as exc_info:
+        if exc_info.returncode == 128:
+            # The file didn't exist at the given revision. Act as if it was an empty
+            # file, so all current lines appear as edited.
+            return []
```
Carreau, line 31: I think you `sys.exit()` in 48, so it will conflict. Probably 48 need to not sys.exit.

Carreau: nevermind you sys.exit in _git_check_output_lines

src/darker/tests/test_main_revision.py
```diff
@@ -0,0 +1,111 @@
...
+    git_repo.add(
+        {"+1M0.py": "MODIFIED=1\n", "+2M1-0.py": None}, commit="Third commit",
+    )
+    # Working tree:
+    for path in paths.values():
+        path.write(worktree_content)
+    arguments = ["--diff", "--revision", revision, "."]
+    should_raise = expect is SystemExit
+
+    with pytest.raises(SystemExit) if should_raise else nullcontext():
```
Carreau, line 102: Ha, I never even considered you could use if expression as a context manager.

akaihola: In the past I've used a custom helper for expressing both values and exceptions as expected outcomes. But this trick I found yesterday really makes it simple.

README.rst
akaihola, line 116:
```suggestion
                           Git revision against which to compare the working
                           tree. Tags, branch names, commit hashes, and other
                           expressions like HEAD~5 work here.
```

src/darker/command_line.py
```diff
@@ -30,6 +30,12 @@ def parse_command_line(argv: List[str]) -> Namespace:
         help="Path(s) to the Python source file(s) to reformat",
         metavar="PATH",
     )
+    parser.add_argument(
+        "-r",
+        "--revision",
+        default="HEAD",
+        help="Git revision against which to compare the working tree",
```
akaihola, line 37: To match your suggestion in the README:
```suggestion
        help=(
            "Git revision against which to compare the working tree. Tags, branch"
            " names, commit hashes, and other expressions like HEAD~5 work here."
        ),
```

README.rst
akaihola: Good idea. I expanded that as a suggestion in this review. Also did the corresponding thing in the argument parser.

Carreau:
I think that `--revision` + `--check`, will be really neet for CI. It will allow project to at least make sure that all contributions past a certain point are compliant, and I'll be able to likely update [one of my bot](https://github.com/MeeseeksBox/MeeseeksDev/blob/34fa2c7d26a0e376b2ff92fbc36c69cfe2ebf187/meeseeksdev/meeseeksbox/commands.py#L114-L150), to actually comment on PRs using the github suggestion to actually automatically fix PRs.

setup.cfg
```diff
@@ -26,6 +26,7 @@ packages = find:
 install_requires =
     black
     typing-extensions ; python_version < "3.8"
+    dataclasses ; python_version < "3.7"
 python_requires = >=3.6
```
Carreau, line 30: You can also considered being 3.7+ only ...

akaihola: Well, Black and isort both still support Python 3.6, and maybe it makes sense to be in sync with them. But it's a good point, we definitely should drop support for old Python versions periodically.

akaihola:
> I was also thinking in a config file (pyproject.toml at the root of the repo?) having the ability to give a revision to be used by default. But +1.

That's a good idea, @Carreau. I created #49 to think about configuration files for Darker. I think we should solve that in a separate PR.

Carreau:
Sorry, approved ! 


> I created #49 to think about configuration files for Darker.

Yes, I saw, and you tagged 1.2.0, which I think is reasonable. Just  `-r` is I think the last thing needed to be able to run darker in ci as you can hardcode the revision in the test script.

PR #44, opened 2020-08-07
Allow to give a commitish reference.

Carreau:
This allow to give a commit-ish reference and can thus work across
multiple commits.

I find this convenient as I usually want to run darker after having
committed or on a pull-request with many changes. I'm also thinking that
for example  if the commit-ish is `origin/master`, or in a config file,
some commit at some point in time; eventually all the lines will be
touched and this will become equivalent to isort + black.

I'm opened to better names for the option.

The PR itself is in 2 commits, second one being the result of `darker  . --commitish HEAD~1` on self.

### DIFF
--- a/src/darker/command_line.py
+++ b/src/darker/command_line.py
@@ -43,7 +43,7 @@ def parse_command_line(argv: List[str]) -> Namespace:
         help="Commit to consider for the diff, default to HEAD (unstaged changes).",
         nargs=1,
         default=[''],
-        metavar='COMMITISH'
+        metavar='COMMITISH',
     )
     parser.add_argument(
         "--check",
--- a/src/darker/command_line.py
+++ b/src/darker/command_line.py
@@ -39,6 +39,13 @@ def parse_command_line(argv: List[str]) -> Namespace:
         help="Don't write the files back, just output a diff for each file on stdout",
     )
     parser.add_argument(
+        "--commitish",
+        help="Commit to consider for the diff, default to HEAD (unstaged changes).",
+        nargs=1,
+        default=[''],
+        metavar='COMMITISH'
+    )
+    parser.add_argument(
         "--check",
         action="store_true",
         help=(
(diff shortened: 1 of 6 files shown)


Carreau:
closing see #43

src/darker/command_line.py
```diff
@@ -38,6 +38,13 @@ def parse_command_line(argv: List[str]) -> Namespace:
         action="store_true",
         help="Don't write the files back, just output a diff for each file on stdout",
     )
+    parser.add_argument(
+        "--commitish",
+        help="Commit to consider for the diff, default to HEAD (unstaged changes).",
+        nargs=1,
```
akaihola, line 44: I learned somthing new: Interestingly, `nargs=1` only means that one token is accepted after the option. Only the last occurrence of a repeated option is considered. So:
```bash
$ darker --commitish one --commitish two file.py
=> args.commitish == ["two"]
$ darker --commitish one --commitish two three file.py
=> args.commitish == ["two"], args.src = ["three", "file.py"]
$ darker --commitish one two --commitish three file.py
darker: error: unrecognized arguments: myfile.py
```

With my solution without `nargs=1`, the behavior is the same, except that we conveniently get a `str` instead of a `List[str]`:
```bash
$ darker --r one --r two file.py
=> args.revision == "two"
$ darker --r one --r two three file.py
=> args.revision == "two", args.src = ["three", "file.py"]
$ darker --r one two --r three file.py
darker: error: unrecognized arguments: myfile.py
```

Really limiting to one repeat requires a extra hack, as illustrated in [this Stack Overflow answer](https://stackoverflow.com/a/23032953/15770).

A repeated argument can be specified with `action="append"`, but there's no way to limit the repeats:
```python
parser.add_argument("-x", action="append", default=["DEFAULT"])
```
```bash
$ myscript.py -x one -x two -x three
=> args.x == ["DEFAULT", "one", "two", "three"]
```

Maybe it would be good to just switch to Click (#47).

PR #48, opened 2020-08-08
Reformat newly added files, too

akaihola:
`git diff --name-only` didn't include newly created untracked files so they were not included in reformatting. A call to `git ls-files` with suitable arguments is now made in addition so untracked files will be included.

### DIFF
--- a/src/darker/git.py
+++ b/src/darker/git.py
@@ -1,8 +1,9 @@
 """Helpers for listing modified files and getting unmodified content from Git"""
 
 import logging
+import sys
 from pathlib import Path
-from subprocess import check_output
+from subprocess import CalledProcessError, check_output
 from typing import Iterable, List, Set
 
 from darker.diff import diff_and_get_opcodes, opcodes_to_edit_linenums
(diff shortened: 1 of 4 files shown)


akaihola:
@Carreau, could you review this one?

src/darker/git.py
```diff
@@ -26,8 +27,24 @@ def should_reformat_file(path: Path) -> bool:
...
-    """Run ``git diff --name-only`` and return file names from the output
+def _git_check_output_lines(cmd: List[str], cwd: Path) -> List[str]:
+    """Log command line, run Git, split stdout to lines, exit with 123 on error"""
+    logger.debug("[%s]$ %s", cwd, " ".join(cmd))
+    try:
+        return check_output(cmd, cwd=str(cwd)).decode("utf-8").splitlines()
+    except CalledProcessError as exc_info:
+        if exc_info.returncode == 128:
+            # Bad revision or another Git failure
+            sys.exit(123)
```
Carreau, line 38: 123 or 128 ? maybe you tried to keep the git error code ?

akaihola: 123 is what Black returns. I thought it's good to stay as similar as possible to it for the benefit of integration with other tools.

src/darker/git.py
```diff
@@ -26,8 +27,24 @@ def should_reformat_file(path: Path) -> bool:
...
-def git_diff_name_only(paths: Iterable[Path], cwd: Path) -> Set[Path]:
-    """Run ``git diff --name-only`` and return file names from the output
+def _git_check_output_lines(cmd: List[str], cwd: Path) -> List[str]:
+    """Log command line, run Git, split stdout to lines, exit with 123 on error"""
+    logger.debug("[%s]$ %s", cwd, " ".join(cmd))
+    try:
+        return check_output(cmd, cwd=str(cwd)).decode("utf-8").splitlines()
+    except CalledProcessError as exc_info:
+        if exc_info.returncode == 128:
+            # Bad revision or another Git failure
```
akaihola, line 37:
```suggestion
            # Bad revision or another Git failure. Follow Black's example and exit with
            # a status code of 123 for an internal error.
```

src/darker/git.py
akaihola: From [Black's README]:
```
  --check                         Don't write the files back, just return the
                                  status.  Return code 0 means nothing would
                                  change.  Return code 1 means some files
                                  would be reformatted. Return code 123 means
                                  there was an internal error.
```

PR #33, opened 2020-07-30
Add support for linters

akaihola:
In addition to Black and Isort, darker gains here the ability to run a linter or linters, and only output linting errors/warnings for edited lines.

This has been tested with
- Pylint (`-L pylint`; requires an upcoming Pylint release with Isort 5 support)
- Mypy (`-L mypy`)
- Flake8 (`-L flake8`)

Probably others work as well – this only depends on the error/warning lines in the output to be in one of the following formats:
- `<path>:<linenum>: <description>`
- `<path>:<linenum>:<column>: <description>`

PR #54, opened 2020-08-14
Bugfix: Linter support to diff against given revision

akaihola:
When rebasing linter support in #33, I missed the need to pass the `-r` / `--revision` option there, too.

With this patch, linter output is now filtered correctly by comparing the working tree against the given revision, instead of against `HEAD`.

Could you @Carreau or @samoylovfp take a look at this quick fix?

### DIFF
--- a/src/darker/linting.py
+++ b/src/darker/linting.py
@@ -58,12 +58,15 @@ def _parse_linter_line(
     return path_in_repo, linenum
 
 
-def run_linter(cmdline: List[str], git_root: Path, paths: Set[Path]) -> None:
+def run_linter(
+    cmdline: List[str], git_root: Path, paths: Set[Path], revision: str
+) -> None:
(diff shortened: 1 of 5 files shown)


src/darker/linting.py
```diff
@@ -75,7 +78,7 @@ def run_linter(cmdline: List[str], git_root: Path, paths: Set[Path]) -> None:
     )
     # assert needed for MyPy (see https://stackoverflow.com/q/57350490/15770)
     assert linter_process.stdout is not None
-    edited_linenums_differ = EditedLinenumsDiffer(git_root)
+    edited_linenums_differ = EditedLinenumsDiffer(git_root, revision)
```
Carreau, line 81: Do you want to make the revision mandatory in EditedLinenumsDiffer ?

akaihola: Thanks, that's indeed a good idea.

PR #57, opened 2020-08-15
Linter highlighting

akaihola:
This also refactors highlighting support into a separate module and uses that in `__main__.py` for `--diff` output highlighting.


akaihola:
Here's example output from running [coverage.py](https://github.com/nedbat/coveragepy/) on the test suite and displaying edited lines with no coverage using [cov_to_lint.py](https://gist.github.com/akaihola/2511fe7d2f29f219cb995649afd3d8d2) as the linter:

```bash
$ cd ~/darker
$ git checkout 1.0.0
$ coverage run -m pytest -c /dev/null
$ darker --revision 0.1.0.. --check --lint cov_to_lint.py src
```
Kuvakaappaus - 2020-12-29 19-09-34

PR #58, opened 2020-08-16
Read configuration from pyproject.toml

akaihola:
Also dumps configuration on terminal if debug level logging is enabled with the `-vv` or `--verbose --verbose` command line option.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,9 +2,6 @@
 # Minimum requirements for the build system to execute.
 requires = ["setuptools", "wheel"]  # PEP 508 specifications.
 
-[tool.black]
-skip-string-normalization = true
-
 [tool.isort]
 multi_line_output = 3
 include_trailing_comma = true
@@ -18,9 +15,3 @@ src = [
     "src",
 ]
 revision = "master"
-isort = true
-lint = [
-    "flake8",
-    "mypy",
-    "pylint",
-]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,3 +12,15 @@ force_grid_wrap = 0
 use_parentheses = true
 line_length = 88
 known_third_party = ["pytest"]
+
+[tool.darker]
+src = [
+    "src",
+]
+revision = "master"
+isort = true
+lint = [
+    "flake8",
+    "mypy",
+    "pylint",
+]
--- a/src/darker/__main__.py
+++ b/src/darker/__main__.py
@@ -6,8 +6,6 @@ from difflib import unified_diff
 from pathlib import Path
 from typing import Generator, Iterable, List, Tuple
 
-import toml
-
 from darker.black_diff import BlackArgs, run_black
 from darker.chooser import choose_lines
 from darker.command_line import ISORT_INSTRUCTION, parse_command_line
--- a/src/darker/__main__.py
+++ b/src/darker/__main__.py
@@ -6,9 +6,12 @@ from difflib import unified_diff
 from pathlib import Path
 from typing import Generator, Iterable, List, Tuple
 
+import toml
+
 from darker.black_diff import BlackArgs, run_black
 from darker.chooser import choose_lines
 from darker.command_line import ISORT_INSTRUCTION, parse_command_line
+from darker.config import dump_config
 from darker.diff import diff_and_get_opcodes, opcodes_to_chunks
 from darker.git import EditedLinenumsDiffer, git_get_modified_files
 from darker.import_sorting import apply_isort, isort
@@ -23,7 +26,7 @@ def format_edited_parts(
     srcs: Iterable[Path],
     revision: str,
     enable_isort: bool,
-    linter_cmdlines: List[List[str]],
+    linter_cmdlines: List[str],
     black_args: BlackArgs,
 ) -> Generator[Tuple[Path, str, str, List[str]], None, None]:
     """Black (and optional isort) formatting for chunks with edits since the last commit
@@ -53,7 +56,7 @@ def format_edited_parts(
     :param revision: The Git revision against which to compare the working tree
     :param enable_isort: ``True`` to also run ``isort`` first on each changed file
     :param linter_cmdlines: The command line(s) for running linters on the changed
-                            files. Each entry is a list of tokens on the command line.
+                            files.
     :param black_args: Command-line arguments to send to ``black.FileMode``
     :return: A generator which yields details about changes for each file which should
              be reformatted, and skips unchanged files.
@@ -192,15 +195,21 @@ def main(argv: List[str] = None) -> int:
     """
     if argv is None:
         argv = sys.argv[1:]
-    args = parse_command_line(argv)
-    log_level = logging.WARNING - sum(args.log_level or ())
-    logging.basicConfig(level=log_level)
-    if log_level == logging.INFO:
+    args, config, config_nondefault = parse_command_line(argv)
+    logging.basicConfig(level=args.log_level)
+    if args.log_level == logging.INFO:
         formatter = logging.Formatter("%(levelname)s: %(message)s")
         logging.getLogger().handlers[0].setFormatter(formatter)
 
     # Make sure we don't get excessive debug log output from Black
     logging.getLogger("blib2to3.pgen2.driver").setLevel(logging.WARNING)
+
+    if args.log_level <= logging.DEBUG:
+        print("\n# Effective configuration:\n")
+        dump_config(config)
+        print("\n# Configuration options which differ from defaults:\n")
+        dump_config(config_nondefault)
+        print("\n")
 
     if args.isort and not isort:
         logger.error(f"{ISORT_INSTRUCTION} to use the `--isort` option.")
--- a/src/darker/__main__.py
+++ b/src/darker/__main__.py
@@ -204,9 +204,9 @@ def main(argv: List[str] = None) -> int:
 
     if args.log_level <= logging.DEBUG:
         print("\n# Effective configuration:\n")
-        dump_config(config)
+        print(dump_config(config))
         print("\n# Configuration options which differ from defaults:\n")
-        dump_config(config_nondefault)
+        print(dump_config(config_nondefault))
         print("\n")
 
     if args.isort and not isort:
--- a/src/darker/black_diff.py
+++ b/src/darker/black_diff.py
@@ -36,7 +36,7 @@ import logging
 import sys
 from functools import lru_cache
 from pathlib import Path
-from typing import Callable, Dict, List, Optional, Set, Union, cast
+from typing import List, Optional, Set, cast
 
 if sys.version_info >= (3, 8):
     from typing import TypedDict
@@ -45,8 +45,7 @@ else:
 
 # `FileMode as Mode` required to satisfy mypy==0.782. Strange.
 from black import FileMode as Mode
-from black import TargetVersion, format_str, read_pyproject_toml
-from click import Command, Context, Option
+from black import TargetVersion, find_pyproject_toml, format_str, parse_pyproject_toml
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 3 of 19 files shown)


akaihola:
I'm drafting [a blog post](https://dev.to/akaihola/incremental-quality-improvement-for-python-projects-2dg9-temp-slug-8732809?preview=0e250a07cb2ce62d507dc3e60c82641be981da80ef445dd93a9423a6c209c583533307878ca220c49ec0aa029d1f8b565033eeef416abc404c938119) about using `darker` and `pytest-darker` together. I realized that the setup I'm going to describe there will require the ability to use a configuration file configure how Darker is run.

akaihola:
@Carreau, are you interested to review this?

(I still can't assign you as a reviewer in the GitHub UI since you haven't accepted the invitation to join as a collaborator. Do you want me to cancel the invitation?)

Carreau:
Sorry I missed the invitation, it has expired now.

Carreau:
did a quick read through, look good to me. I'll look at the blog post.

Carreau:
On the blog post, maybe extend to _why_ one might not want to run darker on that full code base at once. 

Also add that with darker enabled the codebase will thus slowly converge to a point where all code is comformant in the limit of all lines having changed.

src/darker/config.py
```diff
@@ -0,0 +1,62 @@
...
+
+def replace_log_level_name(config: DarkerConfig) -> None:
+    """Replace numeric log level in configuration with the name of the log level"""
+    if "log_level" in config:
+        config["log_level"] = logging.getLevelName(cast(int, config["log_level"]))
+
+
+def load_config() -> DarkerConfig:
+    """Find and load Darker configuration from given path or pyproject.toml"""
+    path = find_project_root((".",)) / "pyproject.toml"
```
akaihola, line 32: Is it confusing that Darker configuration is always read from `pyproject.toml` of the repository, while the `--config` command line option only affects where Black and isort configuration is read from?

akaihola:
@Carreau I did some major fixes after your review. Do you have time to review again, or would you prefer other contributors to join?

Carreau:
> Do you have time to review again, or would you prefer other contributors to join?

I'll try to do that now.

pyproject.toml
```diff
@@ -12,3 +12,15 @@ force_grid_wrap = 0
 use_parentheses = true
 line_length = 88
 known_third_party = ["pytest"]
+
+[tool.darker]
+src = [
+    "src",
+]
+revision = "master"
+isort = true
```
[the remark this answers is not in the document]
Carreau: I think if people ask us to, maybe, but for now I think that's fine.

src/darker/config.py
Carreau: No I think that is fine.

I'm just worried of loading config always wrt `.`, especially from within editors CWD/. may not be what you believe.

src/darker/command_line.py
```diff
@@ -127,4 +139,26 @@ def parse_command_line(argv: List[str]) -> Namespace:
...
+
+    Also return the effective configuration which combines defaults, the configuration
+    read from ``pyproject.toml`` (or the path given in ``--config``), and command line
+    arguments.
+
+    Finally, also return the set of configuration options which differ from defaults.
+
+    """
+    config = load_config()
+    parser = make_argument_parser(require_src=not config.get("src"))
```
Carreau, line 156: I think that should always be required, and that load_config sound take that as a parameter to find the project root.

I think mainly in editors the CWD may not be "in" the project; so relying on global state is a code smell for me.

src/darker/config.py
```diff
@@ -0,0 +1,62 @@
...
+        if value != parser.get_default(argument)
+    }
+    replace_log_level_name(not_default)
+    return not_default
+
+
+def dump_config(config: DarkerConfig) -> None:
+    """Print the configuration in TOML format"""
+    print("[tool.darker]")
+    print(toml.dumps(config, encoder=TomlArrayLinesEncoder()))  # type: ignore[call-arg]
```
Carreau, line 62: Make this return a string and ...

src/darker/__main__.py
```diff
@@ -192,16 +193,22 @@ def main(argv: List[str] = None) -> int:
...
 
     # Make sure we don't get excessive debug log output from Black
     logging.getLogger("blib2to3.pgen2.driver").setLevel(logging.WARNING)
 
+    if args.log_level <= logging.DEBUG:
+        print("\n# Effective configuration:\n")
+        dump_config(config)
+        print("\n# Configuration options which differ from defaults:\n")
+        dump_config(config_nondefault)
+        print("\n")
```
Carreau, line 210: ... add print there ?

Carreau:
Looks good, one suggestion about `print()/dumps_config`, and one concern about making finding project root without a `src`

src/darker/config.py
akaihola: Ah, I suppose you're suggesting that whenever at least one file/directory is given as an argument, the configuration should be searched for going up the directory hierarchy from there?

Carreau: Yes.

Carreau: I'm also worried about what happens w/o arguments. I think that if no argument get passed you likely don't want to _do_ anything and print help. I think it's safer.

My Home is a git repository, so for example running `darker` without a path from within VIM would literally try to reformat all the files in my home directory ...

src/darker/command_line.py
akaihola: What I'm aiming at here is a convenience configuration: if I specify e.g. the main code directory using
```toml
[tool.darker]
src = [
    "src",
]
```
in `pyproject.toml`, then running a plain `darker` command without arguments in that directory should do the right thing and check files in the `src/` tree.

This is somewhat analogous to e.g. `pytest` – according to [documentation](https://docs.pytest.org/en/stable/customize.html#pyproject-toml), in `pyproject.toml` you can write:
```toml
[tool.pytest.ini_options]
testpaths = [
    "tests",
    "integration",
]
```
and running just `pytest` will discover tests only in `tests/` and `integration/`.

What I did here work so that if you _don't_ specify `src = [...]` in `pyproject.toml` for Darker, it _will_ require at least one path on the command line and fail if none are provided.

Carreau: Ah, ok, I misunderstood. Yes, if running from the root of a project, doing `darker` without argument should probably work.

akaihola: ...and let's say I have this:
```
.
├── pyproject.toml
└── src
    └── darker
        ├── __init__.py
        └── __main__.py
```
and this in `pyproject.toml`:
```toml
[tool.darker]
src = [
    "src",
]
```
Then should running `darker .` at the root be equivalent to
- `darker . src`,
- or just `darker src`?

Without `src = [...]` in `pyproject.toml` it's obviously equivalent just to `darker .`.

But anyway, I'll fix it so that `pyproject.toml` is searched for relative to any given paths on the command line. The behavior when no paths are given is debatable, let's iron that out.

akaihola:
Seems like the update to Black 20.8b1 brought a lot of test failures. I wonder if it makes sense to fix those in this branch or separately first in a branch dedicated to them.

akaihola:
> Seems like the update to Black 20.8b1 brought a lot of test failures. I wonder if it makes sense to fix those in this branch or separately first in a branch dedicated to them.

Actually it probably makes sense to fix them here, because it's a perfect test for the configuration file feature! We can add [pytest-darker](https://pypi.org/project/pytest-darker/) to test dependencies, and configure `pytest` to run Darker instead of Black. In `pyproject.toml`, we can specify `src/` as the directory tree to format, and `master` as the revision to compare against.

This way we'll actually converge towards the new Black formatting rules little by little using Darker.

akaihola:
Oh, and another use case for omitting the file path arguments:

[pytest-darker](https://pypi.org/project/pytest-darker/) currently only adds the `pytest --darker` option, and doesn't allow passing command line arguments to Darker. I've figured that the combination of that simple option and supporting Darker configuration in `pyproject.toml` is enough to enable a nice CI setup.

Adding the following files and running the CI script on feature branches in a CI environment should nicely fail only if tests fail, there are formatting, linting or type checking failures on modified lines, or imports are not sorted in modified files:

```ini
;; pytest.ini

[pytest]
addopts =
  --darker
```

```toml
# pyproject.toml

[tool.isort]
multi_line_output = 3
include_trailing_comma = true
force_grid_wrap = 0
use_parentheses = true
line_length = 88

[tool.darker]
src = [
    "src",
]
revision = "master"
isort = true
lint = [
    "flake8",
    "mypy",
    "pylint",
]
```

```ini
;; setup.cfg

[options.extras_require]
test =
    flake8
    mypy
    pylint
    pytest-darker

[flake8]
# Line length according to Black rules
max-line-length = 88
# Ignore rules which conflict with Black
ignore =
    # E231 missing whitespace after ','
    E231
    # W503 line break before binary operator
    W503
```

```bash
# CI script

pip install -e '.[test]'
pytest
```

akaihola:
Ok @Mystic-Mirage now the build in this branch succeeds. We have mixed Black 19.10/20.8 formatting in the code base, and CI only requires Black 20.8 formatting for lines modified in merge requests.

In a way, Darker is now eating its own dogfood :dog: 

Could we merge this instead of #62 in your opinion?

src/darker/black_diff.py
```diff
@@ -64,6 +64,12 @@ class BlackModeAttributes(TypedDict, total=False):
...
 
 
+BLACK_ARG_CONVERSIONS: Dict[str, Callable[[str], Union[int, bool]]] = {
+    "line_length": int,
+    "skip_string_normalization": lambda value: value == "True",
+}
+
+
 @lru_cache(maxsize=1)
 def read_black_config(src: Path, value: Optional[str]) -> BlackArgs:
```
Mystic-Mirage, line 74: I'd propose to rewrite this method using https://github.com/akaihola/darker/pull/62/commits/7bbbf74b761d0ba9163894fcd5c0723e03b198fa#diff-c021797c036b1ba8d6aac676393ca67a

There is no need in explicit type casting for Black args and also we'll finally get rid of Click

akaihola: Thanks @Mystic-Mirage, I cherry picked your fix.

src/darker/config.py
akaihola: >running darker without a path from within VIM would literally try to reformat all the files in my home directory

...but only if you specified `src = ["."]` in your `$HOME/pyproject.toml` file.

src/darker/command_line.py
```diff
@@ -1,13 +1,25 @@
...
+    get_modified_config,
+    load_config,
+)
 from darker.version import __version__
 
 ISORT_INSTRUCTION = "Please run `pip install 'darker[isort]'`"
 
 
-def parse_command_line(argv: List[str]) -> Namespace:
+def make_argument_parser(require_src: bool) -> ArgumentParser:
```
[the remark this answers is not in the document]
akaihola: I think what we have now is enough.

src/darker/config.py
akaihola: > > running darker without a path from within VIM would literally try to reformat all the files in my home directory
> 
> ...but only if you specified `src = ["."]` in your `$HOME/pyproject.toml` file.

@Carreau I hope this satisfies your concern – I'm proceeding to merging this pull request soon.

akaihola:
@Carreau, I continued the [blog post draft](https://dev.to/akaihola/incremental-quality-improvement-for-python-projects-2dg9-temp-slug-8732809?preview=0e250a07cb2ce62d507dc3e60c82641be981da80ef445dd93a9423a6c209c583533307878ca220c49ec0aa029d1f8b565033eeef416abc404c938119) – what do you think about it now?

akaihola:
I've now published the blog post [Improving Python code incrementally](https://dev.to/akaihola/improving-python-code-incrementally-3f7a) on [dev.to](https://dev.to/).

PR #62, opened 2020-08-26
Move to black 20.8b1

Mystic-Mirage:
Black changes its behavior so we need to assume it

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -24,7 +24,7 @@ package_dir =
     =src
 packages = find:
 install_requires =
-    black
+    black>=20.8b1
     typing-extensions ; python_version < "3.8"
     dataclasses ; python_version < "3.7"
 python_requires = >=3.6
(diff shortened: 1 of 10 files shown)


Mystic-Mirage:
@akaihola Could you take a look at test_diff.py and fix it for us? ;-)

akaihola:
Thanks @Mystic-Mirage for this PR.

What I was considering in #58 was:
>Actually it probably makes sense to fix them \[Black 20.8b1 reformatting\] here, because it's a perfect test for the configuration file feature! We can add [pytest-darker](https://pypi.org/project/pytest-darker/) to test dependencies, and configure pytest to run Darker instead of Black. In `pyproject.toml`, we can specify `src/' as the directory tree to format, and master as the revision to compare against.
>
>This way we'll actually converge towards the new Black formatting rules little by little using Darker...

...and that process will serve as a real-life test and proof that Darker (together with pytest-darker) can do what it promises.

akaihola (commented):
I believe #58 addresses the move to Black 20.8b1, and at the same time starts to use Darker as "dogfood" for its own code base. I suggest we merge #58 instead and close this PR.

PR #66, opened 2020-08-28
Support commit ranges for `-r` / `--revision`

akaihola:
This is useful for comparing to the best common ancestor, e.g. `master...`.

I noticed the need for this when I had merges and other commits in `master` after the branch point from which the checked branch started. Without a range, code removed in `master` would get reformatted in the branch.


akaihola:
@CorreyL, would you have time to review these changes?

CorreyL (approved):
Apologies for the delay, looks good to me! A good chunk of it seems to be improving existing documentation, and the small amount of functional changes seems to have an appropriate unit test to go along with it 👍

Issue #60, opened 2020-08-20
Virtual get-together for Darker contributors


akaihola:
Hi @Carreau, @CircleOnCircles, @CorreyL, @DylanYoung, @Mystic-Mirage, @Pacu2, @samoylovfp and anyone else interested in Darker,

As the original author of Darker, I've been super happy to have a bunch of nice contributions from all of you in form of bug reports, suggestions, pull requests, reviews and discussions. Even though it's still obvious that I'm the main driver behind the project, I like to think that there's a nascent community caring for it.

I got this wild idea, probably due to the new on-line lifestyle caused by COVID-19, that it would be nice to have a small virtual get-together between those of us who feel comfortable spending a few moments with each other.

I'm not pushing this in any way, just being curious and probing for interest. Time zones and everyone's work and family schedules are of course a limiting factor, so my hopes aren't too high. But anyway, here goes:

I'm currently in **UTC+3** (East European Standard Time, Daylight Savings). I can easily carve a 30-minute slot from my work and family life during weekdays and weekends **between 5:00 a.m. and 7:00 p.m. UTC**. I prefer to use [meet.jit.si](https://meet.jit.si/) sice it's zero install, but I can also use at least Teams, Slack, Skype, Zoom and Element.io. I'd appreciate if those of you interested in this idea would post your time zones as well as schedule and video call platform limitations, so we can see when and how it's worth trying to schedule a call.

With warmest regards,
Antti

CorreyL:
Happy to connect with other contributors! My time zone is PST, so I could try to make time to be available between **11:00pm and 4:00am UTC** (4pm-9pm my time) during week days, otherwise **1:00pm and 4:00am UTC** on weekends (6am - 9pm my time).

Hope I got those times right, using https://www.worldtimebuddy.com/?pl=1&lid=100,8&h=100

CircleOnCircles:
Hi everyone, I'm too want to join and engage in this meetup.  😀 This could be my first time meet up with a group from open source community.
my preferred time is any day within ** 1:00 am to 10:00 am UTC** Let me know 3 days ahead would be perfect to prevent conflict.

akaihola:
Hey all, thanks for your favorable replies!

If I've intrepreted all correctly, we can get
- weekends 3pm–9pm UTC @akaihola @CorreyL @samoylovfp @Mystic-Mirage @DylanYoung 
[...]

PR #69, opened 2020-09-02
Improve example in README

akaihola:
Also point out in a more clear way that the path arguments can be both files and directories.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -78,11 +78,11 @@ To install, use::
 
   pip install darker
 
-The ``darker <myfile.py>`` command
-reads the original file,
-formats it using black_,
+The ``darker <myfile.py>`` or ``darker <directory>`` command
+reads the original file(s),
+formats them using black_,
 combines original and formatted regions based on edits,
-and writes back over the original file.
+and writes back over the original file(s).
 
 Alternatively, you can invoke the module directly through the ``python`` executable,
 which may be preferable depending on your setup.
@@ -106,22 +106,36 @@ Example:
 
    $ mkdir test && cd test && git init
    Initialized empty Git repository in /tmp/test/.git/
-   $ echo "if True: print('hi')\n\nif False: print('there')" | tee test.py
-   if True: print('hi')
 
+   $ echo "if True: print('hi')\nprint()\nif False: print('there')" | tee test.py
+   if True: print('hi')
+   print()
    if False: print('there')
+
    $ git add test.py && git commit -m "Initial commit"
    [master (root-commit) a0c7c32] Initial commit
     1 file changed, 3 insertions(+)
     create mode 100644 test.py
-   $ echo "if True: print('changed')\n\nif False: print('there')" | tee test.py
-   if True: print('changed')
 
+   $ echo "if True: print('changed')\nprint()\nif False: print('there')" | tee test.py
+   if True: print('changed')
+   print()
    if False: print('there')
-   $ darker test.py && cat test.py
+
+   $ darker --diff .   # show changes in any file under current directory
+   --- test.py
+   +++ test.py
+   @@ -1,3 +1,4 @@
+   -if True: print('changed')
+   +if True:
+   +    print("changed")
+   print()
+   if False: print('there')
+
+   $ darker test.py && cat test.py   # apply changes just to test.py
    if True:
        print("changed")
-
+   print()
    if False: print('there')
 
 Customizing ``darker``, Black and isort behavior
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -106,22 +106,36 @@ Example:
 
    $ mkdir test && cd test && git init
```
markddavidoff, line 107: tbh, this whole example block is not really formatted in a readable way or consistent with the way i see example blocks in other documentation, perhaps i can offer some suggestions

markddavidoff, line 107:
```suggestion
```
Don't really need this, as it is a given that we will have  python files and be in a git directory. instead consider an example showing an error thrown when we are not in a git directory, with a comment re-statating that you need to be in a git directory. Those are often easier to find when people scan the docs while troubleshooting

README.rst
```diff
@@ -106,22 +106,36 @@ Example:
...
    Initialized empty Git repository in /tmp/test/.git/
-   $ echo "if True: print('hi')\n\nif False: print('there')" | tee test.py
-   if True: print('hi')
 
+   $ echo "if True: print('hi')\nprint()\nif False: print('there')" | tee test.py
```
markddavidoff, line 110: For readability, i don't think you should include the code used to generate the examples as they themselves distract away from the actual example. I can't nest backticks in the suggestion block so i'll just make this a comment instead:

### our_file.py at commit hash XXXXXXXX
```
if True: print('hi')

if False: print('there')
```
### An uncommitted change to the our_file.py:
<pre><code>if True: print('<b>changed</b>')

if False: print('there')
</code></pre>

### Show changes to the file(s) darker would make
Note, the `if False: print('there')` is omitted as it was not modified!
```
$ darker --diff ./our_file.py -r XXXXXXXX
> --- test.py
> +++ test.py
> @@ -1,3 +1,4 @@
> -if True: print('changed')
> +if True:
> +    print("changed")
```
You can also specify all files in the current directory:
all files under the current directory
```
$ darker --diff . -r XXXXXXXX
```

### Black the file(s) with darker!
```
$ darker ./our_file.py  -r XXXXXXXX
```
or 
```
$ darker .  -r XXXXXXXX
```
Output our blacked file:
Note, the `if False: print('there')` is not blacked as it was not been modified!
```
$ cat ./our_file.py
> if True: 
>     print('changed')
> if False: print('there')
```

markddavidoff: @akaihola hopefully you find these suggestions useful, if you'd like i can submit them in another pr

README.rst
akaihola: Yes, I would appreciate your improvements!

akaihola: What I intended to do in this example is give the reader a complete set of command lines they can copy-paste on the terminal, and have the exact thing described happen on their own disk.

But you have a good point on troubleshooting. I'll try adding that.

akaihola:
> sure, adding the file(s) helps clarify the current behavior, but i believe darker should also extend black's behavior by letting you pass multiple file paths.

It does!

You can run
```bash
$ darker first.py second.py subdirectory/deep/third.py
```

README.rst
akaihola: Yes, this is very useful! I actually wanted to combine your suggestions immediately wtih some of my own ideas – what do you think of the example now?

akaihola:
I've been actually dreaming about a screencast. It would be fantastic to show an actual coding session in which some parts of a heavily non-black-formatted repository are edited and interactively reformatted in PyCharm or VS Code using Darker.

README.rst
akaihola: One modification to your suggestion was that I didn't want to introduce the `-r` switch in the very first example – for local interactive use I expect the default comparison to `HEAD` to be the sane default most often. It's in the CI environment where `--revision` becomes especially handy.

akaihola:
By the way @markddavidoff if you're interested to contribute more to Darker, I can invite you as a contributor. You may also be interested in #60 – let us know your time zone and schedules there in case you'd like to meet other contributors on-line.

markddavidoff:
@akaihola unfortunately, although interested, I can't honestly say I'd be able to make time for any meaningful contributions. Please feel free to adopt my suggestions as you wish!

akaihola:
@markddavidoff wrote:
>Please feel free to adopt my suggestions as you wish!

Ok thanks for all your feedback Mark!

I added other contributors as additional reviewers – I hope someone check how I modified the example in `README.rst` to give a second opinion and to make sure I didn't miss something.

CorreyL (approved):
Definitely makes what the example was originally trying to get across much easier to read!

PR #83, opened 2020-10-06
Refactor conditional raise assertion in tests

akaihola:
- [x] refactor context manager conditionals into a helper
- [x] fix test failures for Python 3.8
- [x] fix test failures for Python 3.7
- [x] fix test failures for Python 3.6

Some parameterized tests have both values and exceptions as expected outcomes. With recent Pytest and Mypy versions, the type of the conditional context manager doesn't check when written using Python's ternary operator. We get these errors when Pytest runs Mypy checks for Darker:
```python
__________________ src/darker/tests/test_argparse_helpers.py ___________________
33: error: "object" has no attribute "__enter__"
33: error: "object" has no attribute "__exit__"
____________________ src/darker/tests/test_command_line.py _____________________
31: error: "object" has no attribute "__enter__"
31: error: "object" has no attribute "__exit__"
86: error: "object" has no attribute "__enter__"
86: error: "object" has no attribute "__exit__"
____________________ src/darker/tests/test_main_revision.py ____________________
102: error: "object" has no attribute "__enter__"
102: error: "object" has no attribute "__exit__"
```
The problem is that `pytest.raises` and `contextlib.nullcontext` in an `... if ... else ...` ternary operator have no common ancestor classes, so Mypy assigns it the type `object`. To reproduce, create this Python source file:
```python
# ternary_context_manager.py

import pytest
from contextlib import nullcontext

def myfunc(should_raise: bool) -> None:
    cm = pytest.raises(Exception) if should_raise else nullcontext()
    reveal_type(cm)
    with cm:
        pass
```
Analyze it with Mypy in a Python 3.8 virtualenv:
```bash
python3.8 -m venv venv
. venv/bin/activate
pip install pytest==6.1.1 mypy==0.782
mypy ternary_context_manager.py
```
You'll get:
```python
ternary_context_manager.py:8: note: Revealed type is 'builtins.object'
ternary_context_manager.py:9: error: "object" has no attribute "__enter__"
ternary_context_manager.py:9: error: "object" has no attribute "__exit__"
Found 2 errors in 1 file (checked 1 source file)
```
The `with` statement expects the expression to have `__enter__` and `__exit__` methods (i.e. to be a context manager), which `object` of course doesn't have. At run time this isn't a problem, but for Mypy it is.

Fix this by refactoring the conditional into a helper function which returns different context managers using an `if..else` instead. This also reduces duplication of code in tests.

### DIFF
--- a/src/darker/tests/helpers.py
+++ b/src/darker/tests/helpers.py
@@ -1,8 +1,31 @@
 """Helper functions for unit tests"""
 
-from typing import Any, Dict
+import sys
+from typing import Any, ContextManager, Dict, Union
+
+import pytest
+from _pytest.python_api import RaisesContext
+
+if sys.version_info >= (3, 7):
+    from contextlib import nullcontext
+else:
+    from contextlib import suppress as nullcontext
 
 
 def filter_dict(dct: Dict[str, Any], filter_key: str) -> Dict[str, Any]:
     """Return only given keys with their values from a dictionary"""
     return {key: value for key, value in dct.items() if key == filter_key}
+
+
+def raises_if_exception(expect: Any) -> Union[RaisesContext[Any], ContextManager[None]]:
+    """Return a ``pytest.raises`` context manager only if expecting an exception
+
+    If the expected value is not an exception, return a dummy context manager.
+    On Python 3.8+ :class:`contextlib.nullcontext` is returned,
+    while on older Pythons :class:`contextlib.suppress` is used instead.
+
+    """
+    if isinstance(expect, type) and issubclass(expect, BaseException):
+        return pytest.raises(expect)
+    else:
+        return nullcontext()
--- a/setup.cfg
+++ b/setup.cfg
@@ -42,7 +42,7 @@ isort =
     isort>=5.0.1
 test =
     pylint
-    pytest>=5.4
+    pytest>=6.1.0
     pytest-darker
     pytest-flake8>=1.0.6
     pytest-isort>=1.1.0
--- a/src/darker/tests/test_argparse_helpers.py
+++ b/src/darker/tests/test_argparse_helpers.py
@@ -1,12 +1,12 @@
 """Tests for the ``darker.argparse_helpers`` module"""
 
 from argparse import ArgumentParser, Namespace
-from contextlib import suppress
 from logging import CRITICAL, DEBUG, ERROR, INFO, NOTSET, WARNING
 
 import pytest
 
 from darker import argparse_helpers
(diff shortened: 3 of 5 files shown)


CorreyL:
Happy to take a look at this after my shift today!

src/darker/tests/helpers.py
```diff
@@ -1,8 +1,31 @@
...
+
+def raises_if_exception(expect: Any) -> Union[RaisesContext[Any], ContextManager[None]]:
+    """Return a ``pytest.raises`` context manager only if expecting an exception
+
+    If the expected value is not an exception, return a dummy context manager.
+    On Python 3.8+ :class:`contextlib.nullcontext` is returned,
+    while on older Pythons :class:`contextlib.suppress` is used instead.
+
+    """
+    if isinstance(expect, type) and issubclass(expect, BaseException):
```
samoylovfp, line 28: I wonder why this `isinstance` is necessary

samoylovfp (approved):

Mystic-Mirage (approved):
Looks nice

src/darker/tests/helpers.py
CorreyL: I'm curious about the rationale behind this too. My immediate thought is it's a guard clause for `issubclass`, though it'd be good to confirm.

CorreyL (approved):
Changes seem reasonable to me, especially given how much they've tightened up existing tests.

src/darker/tests/helpers.py
akaihola: >guard clause for `issubclass`

Exactly. In parameterized tests, we could be expecting e.g. `int`s or `str`s for some cases, and an exception for others.
```python
>>> issubclass(42, BaseException)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: issubclass() arg 1 must be a class
```

akaihola:
While continuing to experiment with different Python, Pytest and Mypy versions, I expanded the description of this PR with a simple way to reproduce the original situation.

akaihola:
Also Mypy now fails on Python 3.6 and 3.7. Some clever solution is needed in `tests/helpers.py` so Mypy is happy across Python versions.

akaihola:
> Mypy now fails on Python 3.6 and 3.7

Fixed by updating Pytest to >6.1.0 which introduced type hints for the `_pytest.python_api` module.

And by the way, I deem the import from private package `_pytest` acceptable in the test suite. It won't break Darker itself even if that API is changed. We'll only need to fix the test suite if that happens.

CorreyL (approved):

PR #80, opened 2020-09-24
Common ancestor fix

akaihola:
Bugfix: `<commit>...` now compares always correctly to the latest common ancestor

This merge request used to be even larger, but I've managed to trim it down. I hope someone is willing to review it in one go.

**Note:** #107 will fix the build failure.

PR #87, opened 2020-11-01
Add GitRepoFixture helper methods needed by #80

akaihola:
These two additional methods in `darker.tests.conftest.GitRepoFixture` will be needed in unit tests in #80.


CorreyL (approved):
Looks good to me!

PR #90, opened 2020-11-02
test actions

Carreau:


akaihola:
> Sorry, clicked wrong button, did not meant to submit this.

Well, but you seem to have something very interesting cooking there :smirk:

Carreau:
Haha yes, I do have a draft of a github action that post suggested changes back to original PR :-) It's a bit verbose now, so I need to do some stuff like merging changes that are close to each other to avoid having a commit storm, and also find a way to delete previous review when a new commit is pushed; right now for each message deleted this left a "this message has been deleted" ... which is suboptimal; so _maybe_ I should find a way to update previous comments if they are on the same file / range.

Issue #81, opened 2020-10-05
Vim integration docs may want to mention that it requires autoread


shangxiao:
I had to set `autoread` in order to get vim to automatically reload the buffer after applying the autocmd.

PR #82, opened 2020-10-05
docs: Set autoread for vim integration to get the buffer to automatically reload

shangxiao:
Resolves #81

PR #96, opened 2020-11-28
docs: Set autoread for vim integration to get the buffer to automically reload

akaihola:


ivanov (approved):
Just linking up the fact that, if merged, this would close #81 and  #82

Issue #103, opened 2020-12-03
pre-commit hook does not respect `--from-ref` / `--to-ref`


KangOl:
When running darker through `pre-commit`, it ignores the `--from-ref` and `--to-ref` arguments.

These 2 commands should behave the same.
```
pre-commit run darker --from-ref HEAD~1 --to-ref HEAD
darker --revision HEAD~1..HEAD
```

The pre-commit arguments are exposed via the environment variables `PRE_COMMIT_FROM_REF` and `PRE_COMMIT_TO_REF`

akaihola:
Thanks @KangOl for your bug report! We'll definitely fix this for the next release.

Heads up @sherbie @samoylovfp @Pacu2 @CorreyL

PR #109, opened 2020-12-25
Get revision range from pre-commit environment variables

akaihola:
When using the `-r :PRE-COMMIT:` / `--revision=:PRE-COMMIT:` parameter, Darker will expect to find the revision range from the `PRE_COMMIT_FROM_REF` and `PRE_COMMIT_TO_REF` environment variables as set by `pre-commit`.

Based on #108, should be merged after that one.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,7 +1,7 @@
 - id: darker
   name: darker
   description: "Black reformatting to Python files only in regions changed since last commit"
-  entry: darker
+  entry: "darker -r :PRE-COMMIT:"
   language: python
   language_version: python3
   require_serial: true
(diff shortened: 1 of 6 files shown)


akaihola:
@KangOl, does this fix #103 for you?

akaihola:
@sourcery-ai the emojis in the Before / After / Change columns look very similar from a distance due to very similar colors, so it's not easy to spot at a glance which metrics are particularly low or which files changed considerably.

Would some kind of a color scheme make this better?

Also, for my project, the function list table doesn't fit horizontally into the GitHub pull request layout due to some long function names.

brendanator:
@akaihola Thanks for the Sourcery feedback!

PR #120, opened 2021-04-15
Run CI on Windows

martinRenou:
Also changing the config so that the CI runs for all PRs, if that is ok for you.

PR #125, opened 2021-04-23
newline-unit-test-fix

### LINKED ISSUE #124: Newlines are auto-converted on Windows, which causes a test to fail
#120 adds Windows as an environment for the CI pipeline.

This revealed a problem in unit tests for the `TextDocument` class: we take care to detect and keep the newline convention of each file (LF or CR+LF).

While this works great on Unix, on Windows Python seems to already take care of this and convert newlines for us, which [causes unit tests to fail](https://github.com/akaihola/darker/pull/120/checks?check_run_id=2417565210).

Also, it's unclear whether reformatting a CR-only Unix-style would result in a Windows-style CR+LF file.

The original newline handling principle was adopted from Black. Let's look deeper into Black to see how it works on Windows – does it manage to read in files without automatic CR/LF conversion, or does it take that mechanism into account on Windows?

akaihola:
- [x] write source code files in binary in unit tests to prevent newline conversion on Windows
- [x] use a dummy `Path` object instead of `Mock` so a doctest passes correctly on Windows
- [x] also prefer `Path` objects in tests so path separators are correct also on Windows
- [x] use a real file in test when `Path.relative_to()` needs to work – it checks file existence on Windows

This branch sits on top of #120 since we want to see the test result of Windows. Once we see that this fix works, let's rebase this on `master` and merge before merging #120.

### DIFF
--- a/src/darker/black_diff.py
+++ b/src/darker/black_diff.py
@@ -3,8 +3,9 @@
 In examples below, a simple two-line snippet is used.
 The first line will be reformatted by Black, and the second left intact::
 
+    >>> from pathlib import Path
     >>> from unittest.mock import Mock
-    >>> src = Mock()
+    >>> src = Path("dummy/file/path.py")
     >>> src_content = TextDocument.from_lines(
     ...     [
     ...         "for i in range(5): print(i)",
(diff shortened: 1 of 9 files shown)


akaihola:
All tests pass now on Windows, too. I'll rebase this now on `master`.

PR #126, opened 2021-04-24
Return 1 also for linting errors

### LINKED ISSUE #122: Status code returned from mypy checks
When using the --check option alonside mypy, it would be nice if a status of 1 was returned if errors are detected. Maybe we've missed this, but at the moment this makes it quite painful to integrate with CI.

akaihola:
Fixes #122.

- [x] unit test for `main()` with and without linting errors


matclayton (approved):
Generally looks really solid, can't see any issues from my end, but I'm not deeply familiar with the codebase.

PR #134, opened 2021-05-08
Ignore ~/.gitconfig when running git tests

ivanov:
Set HOME environment variable to the temporary directory the tests will
run in, avoiding any ~/.gitconfig settings that may interfere with
assumptions in the tests, particularly init.defaultBranch and
init.templateDir

Context: 
There are two settings that I'm aware of that can violate the assumption that the default branch will be called "master" when running `git init`. The older way is via the templateDir setting, which if pointed to a directory with a HEAD file containing `ref: refs/heads/main` will cause `git init` to default to `main` branch. Additionally, starting with version 2.28, git introduced a defaultBranch setting, that also allows developers to change the name of the initial branch name in a brand new repository.

This PR contains a minimal change to make the git-related tests pass locally (I use the templateDir method to change my default branch name).

Thank you for making and maintaining darker!

---

- [x] review #135 
- [x] merge #135 
- [x] rebase this one on `master`

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -276,10 +276,6 @@ For example:
 
 *New in version 1.2.2:* Support for ``-r :PRE-COMMIT:`` / ``--revision=:PRE_COMMIT:``
 
-*New in version 1.3.0:*
-
-- git-related commands in the test suite now ignore the user's ~/.gitconfig
-
 .. _Black documentation about pyproject.toml: https://black.readthedocs.io/en/stable/pyproject_toml.html
 .. _isort documentation about config files: https://timothycrosley.github.io/isort/docs/configuration/config_files/
 .. _command line arguments: https://black.readthedocs.io/en/stable/installation_and_usage.html#command-line-options
--- a/README.rst
+++ b/README.rst
@@ -276,6 +276,10 @@ For example:
 
 *New in version 1.2.2:* Support for ``-r :PRE-COMMIT:`` / ``--revision=:PRE_COMMIT:``
 
+*New in version 1.3.0:*
+
+- git-related commands in the test suite now ignore the user's ~/.gitconfig
+
 .. _Black documentation about pyproject.toml: https://black.readthedocs.io/en/stable/pyproject_toml.html
 .. _isort documentation about config files: https://timothycrosley.github.io/isort/docs/configuration/config_files/
 .. _command line arguments: https://black.readthedocs.io/en/stable/installation_and_usage.html#command-line-options
--- a/src/darker/tests/conftest.py
+++ b/src/darker/tests/conftest.py
@@ -37,10 +37,11 @@ class GitRepoFixture:
         env = os.environ.copy()
         # for testing, ignore ~/.gitconfig settings like templateDir and defaultBranch
         env["HOME"] = str(root)
-        check_call(["git", "init"], cwd=root, env=env)
-        check_call(["git", "config", "user.email", "ci@example.com"], cwd=root, env=env)
-        check_call(["git", "config", "user.name", "CI system"], cwd=root, env=env)
-        return cls(root, env)
+        instance = cls(root, env)
+        instance._run("init")
+        instance._run("config", "user.email", "ci@example.com")
+        instance._run("config", "user.name", "CI system")
+        return instance
 
     def _run(self, *args: str) -> None:
         """Helper method to run a Git command line in the repository root"""
--- a/src/darker/tests/conftest.py
+++ b/src/darker/tests/conftest.py
@@ -27,23 +27,24 @@ def with_isort():
 
 
 class GitRepoFixture:
-    def __init__(self, root: Path):
+    def __init__(self, root: Path, env: Dict[str, str]):
         self.root = root
+        self.env = env
 
     @classmethod
     def create_repository(cls, root: Path) -> "GitRepoFixture":
         """Fixture method for creating a Git repository in the given directory"""
         env = os.environ.copy()
-        # for testsing, ignore ~/.gitconfig settings like templateDir and defaultBranch
-        env['HOME'] = root
+        # for testing, ignore ~/.gitconfig settings like templateDir and defaultBranch
+        env["HOME"] = str(root)
         check_call(["git", "init"], cwd=root, env=env)
-        check_call(["git", "config", "user.email", "ci@example.com"], cwd=root)
-        check_call(["git", "config", "user.name", "CI system"], cwd=root)
-        return cls(root)
+        check_call(["git", "config", "user.email", "ci@example.com"], cwd=root, env=env)
+        check_call(["git", "config", "user.name", "CI system"], cwd=root, env=env)
+        return cls(root, env)
 
     def _run(self, *args: str) -> None:
         """Helper method to run a Git command line in the repository root"""
-        check_call(["git"] + list(args), cwd=self.root)
+        check_call(["git"] + list(args), cwd=self.root, env=self.env)
 
     def _run_and_get_first_line(self, *args: str) -> str:
         """Helper method to run Git in repo root and return first line of output"""
--- a/src/darker/tests/conftest.py
+++ b/src/darker/tests/conftest.py
@@ -38,6 +38,7 @@ class GitRepoFixture:
         # for testing, ignore ~/.gitconfig settings like templateDir and defaultBranch
         env["HOME"] = str(root)
         instance = cls(root, env)
+        # pylint: disable=protected-access
         instance._run("init")
         instance._run("config", "user.email", "ci@example.com")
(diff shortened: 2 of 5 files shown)


akaihola:
Thanks @ivanov, this is a great catch!

I wonder if it would make sense to also run all the other Git commands besides `git init` this way as well, to prevent other possible side effects from exotic local Git configurations?

src/darker/tests/conftest.py
```diff
@@ -32,7 +33,10 @@ def __init__(self, root: Path):
     @classmethod
     def create_repository(cls, root: Path) -> "GitRepoFixture":
         """Fixture method for creating a Git repository in the given directory"""
-        check_call(["git", "init"], cwd=root)
+        env = os.environ.copy()
+        # for testsing, ignore ~/.gitconfig settings like templateDir and defaultBranch
+        env['HOME'] = root
```
akaihola, line 38:
```suggestion
        env["HOME"] = root
```

Darker's own test pipeline fails if non-Black-compliant code is committed on modified lines. And we don't the `--skip-string-normalization` option of Black anymore.

akaihola:
It would be good to have a unit test which fails without this patch. Also, it's nice to note the improvement in `README.rst`. If you're short on time, I can do those as well.

src/darker/tests/conftest.py
ivanov: How embarrassing that I didn't run darker before submitting a pr to darker  :D 

fixed, thanks!

ivanov:
> I wonder if it would make sense to also run all the other Git commands besides git init this way as well, to prevent other possible side effects from exotic local Git configurations?

Done - now all tests for git commands use env where HOME points to the repo root.

> It would be good to have a unit test which fails without this patch. 

Done - here's what the a904a4305b991803da273fab100884c2fe9a82b7 cherry-picked to master looks like:

```
========================================= FAILURES ===================================
___________________ test_local_gitconfig_ignored_by_gitrepofixture ___________________
tmp_path = PosixPath('/tmp/pytest-of-pi/pytest-29/test_local_gitconfig_ignored_b0')

    def test_local_gitconfig_ignored_by_gitrepofixture(tmp_path):
        """Tests that ~/.gitconfig is ignored when running darker's git tests"""
        with open(tmp_path / "HEAD", "w") as head_file:
            head_file.write("ref: refs/heads/main")

        with patch.dict(os.environ, {"HOME": str(tmp_path)}):
            # Note: once we decide to drop support for git < 2.28, the HEAD file
            # creation above can be removed, and setup can simplify to
            # check_call("git config --global init.defaultBranch main".split())
            check_call("git config --global init.templateDir".split() + [tmp_path])
            root = tmp_path / "repo"
            root.mkdir()
            git_repo = GitRepoFixture.create_repository(root)
>           assert git_repo.get_branch() == "master"
E           AssertionError: assert 'main' == 'master'
E             - master
E             + main

src/darker/tests/test_git.py:467: AssertionError
```

> Also, it's nice to note the improvement in README.rst

I wasn't sure what section to put this in, but hopefully where I landed this was OK. Please feel free to revise or move it elsewhere as you see fit :)

README.rst
```diff
@@ -276,6 +276,10 @@ For example:
 
 *New in version 1.2.2:* Support for ``-r :PRE-COMMIT:`` / ``--revision=:PRE_COMMIT:``
 
+*New in version 1.3.0:*
+
+- git-related commands in the test suite now ignore the user's ~/.gitconfig
+
```
akaihola, line 282: Could you move this to `CHANGES.rst`?

src/darker/tests/test_git.py
```diff
@@ -449,3 +449,19 @@ def test_edited_linenums_differ_revision_vs_lines(git_repo, context_lines, expec
     result = differ.revision_vs_lines(Path("a.py"), content, context_lines)
 
     assert result == expect
+
+
+def test_local_gitconfig_ignored_by_gitrepofixture(tmp_path):
+    """Tests that ~/.gitconfig is ignored when running darker's git tests"""
+    with open(tmp_path / "HEAD", "w") as head_file:
+        head_file.write("ref: refs/heads/main")
```
akaihola, line 457: This could be written as
```suggestion
    (tmp_path / "HEAD").write_text("ref: refs/heads/main")
```

src/darker/tests/test_main.py
```diff
@@ -14,12 +13,13 @@
...
 import darker.import_sorting
 from darker.git import RevisionRange
+from darker.tests.conftest import GitRepoFixture
 from darker.utils import TextDocument
 from darker.verification import NotEquivalentError
 
 
 def test_isort_option_without_isort(tmpdir, without_isort, caplog):
-    check_call(["git", "init"], cwd=tmpdir)
+    GitRepoFixture.create_repository(tmpdir)
```
akaihola, line 22: Why not just use the `git_repo` fixture directly? It does exactly this (plus chdirs into the repo root).
```python
def test_isort_option_without_isort(git_repo, without_isort, caplog):
    # check_call(["git", "init"], cwd=tmpdir)  # remove this...
    # GitRepoFixture.create_repository(tmpdir)  # ...as well as this
```
The `GitRepoFixture` import should then be removed as well.

akaihola (changes requested):
Sorry for the delay @ivanov! Here are a couple of comments still. Let me know if you're busy and would like me to apply these tweaks.

akaihola:
Build fails due to Black having updated to 21.x. See #135 for a fix. Let's merge that one first and rebase.

src/darker/tests/conftest.py
```diff
@@ -26,20 +27,25 @@ def with_isort():
...
         """Fixture method for creating a Git repository in the given directory"""
-        check_call(["git", "init"], cwd=root)
-        check_call(["git", "config", "user.email", "ci@example.com"], cwd=root)
-        check_call(["git", "config", "user.name", "CI system"], cwd=root)
-        return cls(root)
+        env = os.environ.copy()
+        # for testing, ignore ~/.gitconfig settings like templateDir and defaultBranch
+        env["HOME"] = str(root)
+        instance = cls(root, env)
+        instance._run("init")
```
akaihola, line 41: PyLint complains and fails our build unless we explicitly allow access to a protected method of another class instance:
```suggestion
        # pylint: disable=protected-access
        instance._run("init")
```

akaihola (changes requested):

src/darker/tests/test_git.py
ivanov: TIL. old habits die hard.

src/darker/tests/test_main.py
ivanov: sure, ended up having to add a docstring to the test and `# pylint: disable=unused-argument` since neither `git_repo` nor `without_isort` end up getting used.

src/darker/tests/conftest.py
ivanov: I added this, too, but for what it's worth, didn't see a complaint when running the test suite locally or via `darker -L pylint`  with black version 21.5b0 on Python 3.7.3.

ivanov:
Addressed the review comments, hopefully we're all good here, @akaihola , so back over to you.

src/darker/tests/conftest.py
akaihola: Strange, I did see the complaint. Maybe we have different versions of Pylint.

src/darker/tests/test_git.py
```diff
@@ -449,3 +449,18 @@ def test_edited_linenums_differ_revision_vs_lines(git_repo, context_lines, expec
...
+
+def test_local_gitconfig_ignored_by_gitrepofixture(tmp_path):
+    """Tests that ~/.gitconfig is ignored when running darker's git tests"""
+    (tmp_path / "HEAD").write_text("ref: refs/heads/main")
+
+    with patch.dict(os.environ, {"HOME": str(tmp_path)}):
+        # Note: once we decide to drop support for git < 2.28, the HEAD file
+        # creation above can be removed, and setup can simplify to
+        # check_call("git config --global init.defaultBranch main".split())
+        check_call("git config --global init.templateDir".split() + [tmp_path])
```
akaihola, line 462: It seems this needs to be turned into a string – py36 and py37 Windows builds are failing.
```suggestion
        check_call("git config --global init.templateDir".split() + [str(tmp_path)])
```

akaihola (changes requested):
A Windows build failure needs to be addressed.

src/darker/tests/test_git.py
ivanov: Thanks for this!

akaihola (approved):

akaihola:
Thanks @ivanov for this patch!

PR #94, opened 2020-11-28
Binary search to minimize number of context lines

akaihola:
If the AST differs after reformatting when diffing and chunking with zero context lines, search for the lowest successful number of context lines using a binary search instead of by incrementing one by one. This vastly improves performance in such problematic cases.

PR #138, opened 2021-05-16
Fix AST verification and historical comparisons

akaihola:
This pull request fixes three bugs:

**AST verification** no longer erroneously fails when using `--isort`.

This bug was introduced in #94 and is present in version 1.2.3.

**Historical comparisons**, e.g. comparing two past commits like `darker --diff v1.0..v1.1` would still compare against the file on disk in the working tree, not the second commit. This has been fixed.

**Darker would write historical reformats on disk**, overwriting a newer version in the working tree. Now an exception is raised instead if a historical comparison like `darker --revision=commit1..commit2` is used without `--diff` or `--check`.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -10,8 +10,9 @@ Fixed
 -----
 - Git-related commands in the test suite now ignore the user's ``~/.gitconfig``.
 - Now works again even if ``isort`` isn't installed
-- AST verification no longer erroneously fails when using ``--isort`` and a modified
-  Python source code file is already properly formatted
+- AST verification no longer erroneously fails when using ``--isort``
+- Historical comparisons like ``darker --diff --revision=v1.0..v1.1`` now actually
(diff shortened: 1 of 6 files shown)


akaihola:
@samoylovfp, I assigned this to you since we collaborated also on #94 and you may remember the change I made there.

The description isn't actually accurate currently, I'm trying to rethink this through to figure out what's happening exactly.

akaihola:
@ivanov, I added you as a reviewer just in case you'd be interested – you mentioned this PR in the discussion of #140.

akaihola:
@ivanov and/or @samoylovfp, do you think you'll be able to review this?

One option I'd like to try at some point in this project is to have a pair/group video call session where a pull request author walks through their proposed changes. Maybe this PR is a candidate for such?

akaihola:
I re-reviewed this myself. Merging now since other contributors are busy.

PR #139, opened 2021-05-17
Fix assumption of root being CWD in print_diff

ivanov:
As currently written, `print_diff` assumes that that current working directory should be used as the relative path, which does not necessarily need to be the case.

Submitting the test first, for that satisfying red X to green check mark, the patch follows.

- [x] write a test that captures current bug
- [x] make it pass
- [x]  document the change

### DIFF
--- a/src/darker/__main__.py
+++ b/src/darker/__main__.py
@@ -6,7 +6,7 @@ from argparse import Action, ArgumentError
 from datetime import datetime
 from difflib import unified_diff
 from pathlib import Path
-from typing import Collection, Generator, List, Tuple
+from typing import Collection, Generator, List, Optional, Tuple
 
 from darker.black_diff import (
     BlackConfig,
@@ -196,17 +196,22 @@ def modify_file(path: Path, new_content: TextDocument) -> None:
     path.write_bytes(new_content.encoded_string)
 
 
-def print_diff(path: Path, old: TextDocument, new: TextDocument) -> None:
+def print_diff(
+    path: Path, old: TextDocument, new: TextDocument, root: Optional[Path] = None
+) -> None:
     """Print ``black --diff`` style output for the changes
 
     :param path: The relative path of the file to print the diff output for
     :param old: Old contents of the file
     :param new: New contents of the file
+    :param root: The root of the repository (current working directory if omitted)
 
     Modification times should be in the format "YYYY-MM-DD HH:MM:SS:mmmmmm +0000"
 
     """
-    relative_path = path.resolve().relative_to(Path.cwd()).as_posix()
+    if root is None:
+        root = Path.cwd()
+    relative_path = path.resolve().relative_to(root).as_posix()
     diff = "\n".join(
         line.rstrip("\n")
         for line in unified_diff(
@@ -349,7 +354,7 @@ def main(argv: List[str] = None) -> int:
     ):
         failures_on_modified_lines = True
         if output_mode == OutputMode.DIFF:
-            print_diff(path, old, new)
+            print_diff(path, old, new, git_root)
         elif output_mode == OutputMode.CONTENT:
             print_source(new)
         if write_modified_files:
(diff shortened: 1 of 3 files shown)


src/darker/__main__.py
```diff
@@ -136,9 +136,13 @@ def modify_file(path: Path, new_content: TextDocument) -> None:
     path.write_bytes(new_content.encoded_string)
 
 
-def print_diff(path: Path, old: TextDocument, new: TextDocument) -> None:
+def print_diff(
+    path: Path, old: TextDocument, new: TextDocument, root: Optional[Path] = None
```
akaihola, line 140: For keyword arguments with a default value of `None`, actually `Optional[...]` can be omitted and it is assumed by Mypy due to  the default value.

That said, I'm not sure if it's better to write `root: Path = None` or `root: Optional[Path] = None`. What do you think?

src/darker/tests/test_main.py
```diff
@@ -261,14 +267,22 @@ def test_main(
...
+    tmp_path_factory,
 ):  # pylint: disable=too-many-arguments
     """Main function outputs diffs and modifies files correctly"""
-    monkeypatch.chdir(git_repo.root)
+    if root_as_cwd:
+        cwd = git_repo.root
+        pwd = ""
+    else:
+        cwd = tmp_path_factory.mktemp("not_a_git_repo")
+        pwd = str(git_repo.root) + "/"
```
akaihola, line 279: In Darker I've tried to consistently concatenate path elements using `pathlib.Path` methods instead of string operations. So here, too we could do instead:
```suggestion
    if root_as_cwd:
        cwd = git_repo.root
        pwd = Path("")
    else:
        cwd = tmp_path_factory.mktemp("not_a_git_repo")
        pwd = git_repo.root
```

A corresponding change is needed below for line 285.

src/darker/tests/test_main.py
```diff
@@ -261,14 +267,22 @@ def test_main(
...
+    monkeypatch.chdir(cwd)
     paths = git_repo.add({"a.py": newline, "b.py": newline}, commit="Initial commit")
     paths["a.py"].write_bytes(newline.join(A_PY).encode("ascii"))
     paths["b.py"].write_bytes("print(42 ){newline}".encode("ascii"))
 
-    retval = darker.__main__.main(arguments + ['a.py'])
+    retval = darker.__main__.main(arguments + [pwd + "a.py"])
```
akaihola, line 285: Here, a `pathlib.Path` concatenation can be used:
```suggestion
    retval = darker.__main__.main(arguments + [str(pwd  / "a.py")])
```

I've avoided many Windows compatibility gotchas this way, although in this particular case the forward slash didn't break anything in the Windows build. But maybe it's good to do this for consistency.

src/darker/__main__.py
```diff
@@ -4,7 +4,7 @@
 import sys
 from difflib import unified_diff
 from pathlib import Path
-from typing import Generator, Iterable, List, Tuple
+from typing import Generator, Iterable, List, Optional, Tuple
```
akaihola, line 7: This import can be omitted if the shorter form `root: Path = None` is used below in `print_diff()` signature.
```suggestion
from typing import Generator, Iterable, List, Tuple
```

akaihola (changes requested):
The patch looks good (with just minor comments), but it's still unclear to me what is the specific command line use case where things would go wrong without it?

akaihola:
@ivanov, could you rebase your branch on `master` and address my questions in the review? Thanks!

src/darker/__main__.py
akaihola, line 140: For keyword arguments with a default value of `None`, actually `Optional[...]` can be omitted and it is assumed by Mypy due to  the default value.

That said, I'm not sure if it's better to write `root: Path = None` or `root: Optional[Path] = None`. What do you think?

```suggestion
def print_diff(
    path: Path, old: TextDocument, new: TextDocument, root: Path = None
```

akaihola (approved):
@ivanov, I applied my suggested fixes and rebased on `master`. This will be included in release 1.3.1. Thanks for the fix!

Issue #137, opened 2021-05-16
AST verification fails if a modified file is perfectly formatted


akaihola:
It seems that if a Python source code file
- was edited,
- the formatting is already Black compliant,
- imports are already sorted according to `isort` rules, and
- the `--isort` option is used,

then Darker will erroneously terminate with a `darker.verification.NotEquivalentError` exception.

To reproduce, run this in Darker's own repository:
```python
$ darker --diff --isort -r cab1870^..cab1870
Traceback (most recent call last):
  File "bin/darker", line 33, in <module>
    sys.exit(load_entry_point('darker', 'console_scripts', 'darker')())
  File "darker/__main__.py", line 223, in main
    for path, old, new in format_edited_parts(
  File "darker/__main__.py", line 121, in format_edited_parts
    raise NotEquivalentError(path_in_repo)
darker.verification.NotEquivalentError: darker/tests/conftest.py
```
In commit cab1870, a function and an import were removed, the resulting file is already properly formatted, and for sure there aren't any AST differences between the original edited file and the identical reformatted file.

PR #140, opened 2021-05-17
Allow running on plain (non-git) directories

### LINKED ISSUE #71: Improve error when not run in a Git repository
```bash
$ mkdir /tmp/test ; cd /tmp/test ; darker .
error: Could not access 'HEAD'
Traceback (most recent call last):
  File "/home/akaihola/.virtualenvs/darker/bin/darker", line 33, in <module>
    sys.exit(load_entry_point('darker', 'console_scripts', 'darker')())
  File "/home/akaihola/prg/darker/src/darker/__main__.py", line 230, in main
    for path, old_content, new_content, new_lines in format_edited_parts(
  File "/home/akaihola/prg/darker/src/darker/__main__.py", line 64, in format_edited_parts
    changed_files = git_get_modified_files(srcs, revrange, git_root)
  File "/home/akaihola/prg/darker/src/darker/git.py", line 174, in git_get_modified_files
    lines = _git_check_output_lines(diff_cmd, cwd)
  File "/home/akaihola/prg/darker/src/darker/git.py", line 130, in _git_check_output_lines
    return check_output(cmd, cwd=str(cwd)).decode("utf-8").splitlines()
  File "/usr/lib/python3.8/subprocess.py", line 411, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/usr/lib/python3.8/subprocess.py", line 512, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['git', 'diff', '--name-only', '--relative', 'HEAD', '--', '.']' returned non-zero exit status 1.
```

ivanov:
I started off writing some code for better messages when running `darker` on a non-git repo, to fix #71, but then ended up writing enough to make it work anyway.

With this PR, you can `mkdir /tmp/test ; cd /tmp/test ; darker .` and no compaints are raised. Same goes for

```diff
$ mkdir -p /tmp/test ; cd /tmp/test ; echo "print   ('hello'  )" > a.py; darker --diff .;
--- a.py
+++ a.py
@@ -1 +1 @@
-print   ('hello'  )
+print("hello")
```

Opening this up for discussion to see if this if @akaihola would be amenable to something like this, before I start adding tests. In particular,  this will conflict with #138, though I also ran into AST verification failing when using `--isort`, my fix is slightly different (perhaps wrong?). 

#139 also came out of the work on this, but I decided to submit that one separately, since it's an independent change.

### DIFF
--- a/src/darker/__main__.py
+++ b/src/darker/__main__.py
@@ -24,11 +24,11 @@ from darker.git import (
     PRE_COMMIT_FROM_TO_REFS,
     WORKTREE,
     EditedLinenumsDiffer,
-    NotGitRespository,
     RevisionRange,
     get_missing_at_revision,
     git_get_content_at_revision,
     git_get_modified_python_files,
+    git_is_repository,
 )
 from darker.help import ISORT_INSTRUCTION
 from darker.highlighting import colorize
@@ -370,11 +370,11 @@ def main(argv: List[str] = None) -> int:
         black_exclude = set()
     else:
         # In other modes, only process files which have been modified.
-        try:
+        if git_is_repository(root):
             changed_files_to_process = git_get_modified_python_files(
                 files_to_process, revrange, root
             )
-        except NotGitRespository:
+        else:
             changed_files_to_process = {
                 path.relative_to(root) for path in files_to_process
             }
--- a/src/darker/__main__.py
+++ b/src/darker/__main__.py
@@ -1,5 +1,6 @@
 """Darker - apply black reformatting to only areas edited since the last commit"""
 
+import glob
 import logging
 import sys
 import warnings
@@ -24,6 +25,7 @@ from darker.git import (
     PRE_COMMIT_FROM_TO_REFS,
     WORKTREE,
     EditedLinenumsDiffer,
+    NotGitRespository,
     RevisionRange,
     get_missing_at_revision,
     git_get_content_at_revision,
@@ -369,10 +371,18 @@ def main(argv: List[str] = None) -> int:
         black_exclude = set()
     else:
         # In other modes, only process files which have been modified.
-        # These are relative to `root`:
-        changed_files_to_process = git_get_modified_files(
-            files_to_process, revrange, root
-        )
+        try:
+            changed_files_to_process = git_get_modified_files(
+                files_to_process, revrange, root
+            )
+        except NotGitRespository:
+            changed_files_to_process = set()
+            for path in files_to_process:
+                if str(path).endswith(".py"):
+                    changed_files_to_process.add(Path(path))
+                else:
+                    more_files = glob.glob(str(path) + "/**/*.py", recursive=True)
+                    changed_files_to_process.update({Path(p) for p in more_files})
         black_exclude = {
             f for f in changed_files_to_process if root / f not in files_to_blacken
         }
--- a/src/darker/__main__.py
+++ b/src/darker/__main__.py
@@ -28,7 +28,7 @@ from darker.git import (
     RevisionRange,
     get_missing_at_revision,
     git_get_content_at_revision,
-    git_get_modified_files,
+    git_get_modified_python_files,
 )
 from darker.help import ISORT_INSTRUCTION
 from darker.highlighting import colorize
@@ -371,16 +371,13 @@ def main(argv: List[str] = None) -> int:
     else:
         # In other modes, only process files which have been modified.
         try:
-            changed_files_to_process = git_get_modified_files(
+            changed_files_to_process = git_get_modified_python_files(
                 files_to_process, revrange, root
             )
         except NotGitRespository:
-            changed_files_to_process = set()
-            for path in files_to_process:
-                if path.suffix == ".py":
-                    changed_files_to_process.add(path)
-                else:
-                    changed_files_to_process.update(path.glob("**/*.py"))
+            changed_files_to_process = {
+                path.relative_to(root) for path in files_to_process
+            }
         black_exclude = {
             f for f in changed_files_to_process if root / f not in files_to_blacken
         }
--- a/src/darker/__main__.py
+++ b/src/darker/__main__.py
@@ -1,6 +1,5 @@
 """Darker - apply black reformatting to only areas edited since the last commit"""
 
-import glob
 import logging
 import sys
 import warnings
@@ -378,11 +377,10 @@ def main(argv: List[str] = None) -> int:
         except NotGitRespository:
             changed_files_to_process = set()
             for path in files_to_process:
-                if str(path).endswith(".py"):
-                    changed_files_to_process.add(Path(path))
+                if path.suffix == ".py":
+                    changed_files_to_process.add(path)
                 else:
-                    more_files = glob.glob(str(path) + "/**/*.py", recursive=True)
-                    changed_files_to_process.update({Path(p) for p in more_files})
+                    changed_files_to_process.update(path.glob("**/*.py"))
         black_exclude = {
             f for f in changed_files_to_process if root / f not in files_to_blacken
         }
--- a/src/darker/git.py
+++ b/src/darker/git.py
@@ -8,7 +8,7 @@ from dataclasses import dataclass
 from datetime import datetime
 from functools import lru_cache
 from pathlib import Path
-from subprocess import DEVNULL, PIPE, CalledProcessError, check_output, run
+from subprocess import DEVNULL, PIPE, CalledProcessError, run
 from typing import Iterable, List, Set, Tuple
 
 from darker.diff import diff_and_get_opcodes, opcodes_to_edit_linenums
@@ -30,6 +30,10 @@ COMMIT_RANGE_RE = re.compile(r"(.*?)(\.{2,3})(.*)$")
 #   for determining the revision range
 WORKTREE = ":WORKTREE:"
 PRE_COMMIT_FROM_TO_REFS = ":PRE-COMMIT:"
+
+
+class NotGitRespository(Exception):
+    "Raised when git commands are run in a folder that is not a git repository"
 
 
 def git_get_mtime_at_commit(path: Path, revision: str, cwd: Path) -> str:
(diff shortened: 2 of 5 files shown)


src/darker/__main__.py
```diff
@@ -219,7 +230,16 @@ def main(argv: List[str] = None) -> int:
...
+    try:
+        changed_files = git_get_modified_files(paths, revrange, git_root)
+    except NotGitRespository:
+        changed_files = set()
+        for path in paths:
+            if str(path).endswith(".py"):
+                changed_files.add(Path(path))
+            else:
+                more_files = glob.glob(str(path) + "/**/*.py", recursive=True)
+                changed_files.update({Path(p) for p in more_files})
```
ivanov, line 242: We can probably throw this logic into a separate `get_relevant_files_without_git` function, and keep the logic as 

```python
     try:
        changed_files = git_get_modified_files(paths, revrange, git_root)
    except NotGitRespository:
        changed_files = get_relevant_files_without_git(paths)
```

open to other names, too? Maybe `nogit_get_relevant_files`?

src/darker/__main__.py
```diff
@@ -74,6 +80,11 @@ def format_edited_parts(
             )
             if enable_isort and not edited_linenums and edited == worktree_content:
                 logger.debug("No changes in %s after isort", src)
+                last_successful_reformat = (
+                    src,
+                    worktree_content,
+                    TextDocument.from_lines(""),
+                )
```
ivanov, line 87: this was my fix for the false AST non-equivalence thing described in #137, though I make the `chosen` lines empty.

akaihola:
Thanks @ivanov for working on this proposal.

I'm just wondering what the use case for Darker in a non-version-controlled source code tree is. You could just as well run Black, isort and any linters directly on the whole source tree since there are no known "modifications" to limit output to. Or am I misunderstanding your idea?

Is this just for convenience so you can use the same toolchain to do a complete source tree reformatting as you'd use for Git repositories?

ivanov:
> Is this just for convenience so you can use the same toolchain to do a complete source tree reformatting as you'd use for Git repositories?

Yes, this, exactly. I'm just thinking of adding 'darker' to my muscle memory (or some alias for it), and it'd be a shame to have to use an alternate command just because it's not in a git folder. You also gain the the ability to see cumulative diff of the choice of black/isort/pylint that you specify, which would take effort without darker.

akaihola:
>it'd be a shame to have to use an alternate command just because it's not in a git folder.

Thinking deeper into this, you're right, it's a valid use case. I'll proceed to review!

src/darker/__main__.py
akaihola: An empty `chosen` would indicate that all lines have been deleted.

The fix is going to be merged in #138. Could you wait with this PR and rebase on `master` after that?

ivanov: > An empty chosen would indicate that all lines have been deleted.
Ah, ok, I flipped the "sign" in my head - I thought it was the lines that have changed, so an empty chosen would have been from no changes. Thanks for setting me straight :).

> The fix is going to be merged in #138. Could you wait with this PR and rebase on master after that?

Yep, will do that.

src/darker/__main__.py
akaihola: Actually, `git_get_modified_files()` should be called `git_get_modified_python_files()`, and then along the same lines, the new function could be `nogit_get_modified_python_files()`.

src/darker/git.py
```diff
@@ -130,8 +134,13 @@ def _git_check_output_lines(
     """Log command line, run Git, split stdout to lines, exit with 123 on error"""
     logger.debug("[%s]$ %s", cwd, " ".join(cmd))
     try:
-        return check_output(["git"] + cmd, cwd=str(cwd)).decode("utf-8").splitlines()
+        result = run(["git"] + cmd, cwd=str(cwd), check=True, stdout=PIPE, stderr=PIPE)
+        return result.stdout.decode("utf-8").splitlines()
```
akaihola, line 138: Why this change?

src/darker/git.py
```diff
@@ -130,8 +134,13 @@ def _git_check_output_lines(
...
     except CalledProcessError as exc_info:
+        if exc_info.returncode == 129 and exc_info.stderr.decode("utf-8").startswith(
+            "Not a git repository"
+        ):
+            raise NotGitRespository(str(cwd) + " is not a git repository") from exc_info
```
akaihola, line 143: For consistency, I'd use an f-string here like in other corresponding places:
```suggestion
            raise NotGitRespository(f"{cwd} is not a git repository") from exc_info
```

src/darker/__main__.py
```diff
@@ -219,7 +230,16 @@ def main(argv: List[str] = None) -> int:
...
     failures_on_modified_lines = False
     revrange = RevisionRange.parse(args.revision)
-    changed_files = git_get_modified_files(paths, revrange, git_root)
+    try:
+        changed_files = git_get_modified_files(paths, revrange, git_root)
+    except NotGitRespository:
+        changed_files = set()
+        for path in paths:
+            if str(path).endswith(".py"):
+                changed_files.add(Path(path))
```
akaihola, line 239: Since `path` is guaranteed to be a `Path`, this could be simplified to
```suggestion
            if path.suffix == ".py":
                changed_files.add(path)
```

src/darker/__main__.py
akaihola, line 242: This could be simplified too:
```suggestion
                changed_files.update(path.glob("**/*.py"))
```

akaihola (changes requested):

akaihola (changes requested):
We also need unit test cases for non-Git-repository situations, as well as an end-to-end test similar to `darker.tests.test_main.test_main()`.

akaihola:
Hi @ivanov,

I wanted to work on some of my suggestions above, and the only way I could find to do that was to copy this branch to the upstream repository as [akaihola:non-git-repos](https://github.com/akaihola/darker/tree/non-git-repos). 

Do you know if it's possible to switch the branch for this merge request from `ivanov:non-git-repos` to `akaihola:non-git-repos`, or is there a better workflow for doing this kind of a takeover?

Or would you prefer to reset `ivanov:non-git-repos` to point to the latest commit in `akaihola:non-git-repos`?

akaihola:
Ping @ivanov,

I'm preparing a Darker 1.3.0 release. Let me know if you'd like to get this patch in and can spend a few moments on my questions above.

Cheers,
Antti

akaihola:
Ping @ivanov,

I still haven't released 1.3.0. Would you like to get this feature in for the release?

akaihola:
Postponing this feature to at least version 1.3.1.

src/darker/git.py
```diff
@@ -149,23 +153,32 @@ def _git_check_output_lines(
...
-        ).splitlines()
+        )
+        return result.stdout.splitlines()
     except CalledProcessError as exc_info:
+        retval = exc_info.returncode
+        msg = exc_info.stderr
+        if (retval == 129 and msg.startswith("Not a git repository")) or (
+            retval == 1 and msg.startswith("error: could not access ")
+        ):
+            raise NotGitRespository(f"{cwd} is not a git repository") from exc_info
```
akaihola, line 172: I wonder if this is a good way to detect non-Git-directories. The `could not access 'HEAD'` message came up when tests were run on GitHub, but I can't reproduce it locally. We don't know whether that error might occur within Git repositories under different conditions, and this might mask such situations.

akaihola: ...or could there be an alternative explanation why Git would terminate with `error: Could not access 'HEAD'` than a different Git version?

Pytest temporary directories being created in a directory tree which is a Git repository? Sounds unlikely to me.

akaihola: No, this is indeed most probably a difference in Git.

In a Debian 11 Docker image:
```
# git --version
git version 2.30.2
# git diff --name-only --relative HEAD -- .
error: Could not access 'HEAD'
```

Same on Debian Buster:
```
# git --version
git version 2.20.1
root@6c76cb727ad0:/# git diff --name-only --relative HEAD -- .
Not a git repository
To compare two paths outside a working tree:
usage: git diff [--no-index] <path> <path>
```

Same on Fedora 30
```
# git --version
git version 2.21.3
# git diff --name-only --relative HEAD -- .
Not a git repository
To compare two paths outside a working tree:
usage: git diff [--no-index] <path> <path>
```

akaihola: For a non-existing revision, I get this on Git 2.30.2:
```
# git diff --name-only --relative foo -- .
fatal: bad revision 'foo'
```

So at least that won't be confused with a non-Git-directory.

akaihola: Maybe this would be a more reliable detection:

Git 2.21.3 and 2.30.2:
```
# git rev-parse --is-inside-work-tree
true
```

Git 2.21.3:
```
# git rev-parse --is-inside-work-tree
fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
```

Git 2.30.2:
```
# git rev-parse --is-inside-work-tree
fatal: not a git repository (or any of the parent directories): .git
```

akaihola: Oops, there really is something funky going on in the GitHub Actions environment!

In `test_main_in_plain_directory(tmp_path, ...)`, when `main()` calls `git_is_repository()`, it gets `True` as the result even though we should be in a clean temporary directory!

See [this build](https://github.com/akaihola/darker/pull/140/checks?check_run_id=3645005237) for details.

akaihola: **HEISENBUG!**

I added this test in `test_git.py`:
```python
def test_tmp_path_sanity(tmp_path):
    """Make sure Pytest temporary directories aren't inside a Git repository"""
    try:
        result = git._git_check_output_lines(
            ["rev-parse", "--absolute-git-dir"], tmp_path, exit_on_error=False
        )
    except CalledProcessError as exc_info:
        if exc_info.returncode != 128 or not exc_info.stderr.startswith(
            "fatal: not a git repository"
        ):
            raise
    else:
        output = "\n".join(result)
        raise AssertionError(
            f"Temporary directory {tmp_path} for tests is not clean."
            f" There is a Git directory in {output}"
        )
```

And now all tests pass! How did this change temporary directories in other tests?!?

akaihola:
@ivanov, I'm going to put this feature maybe already in version 1.3.1. Would you still like to review my changes?

Issue #133, opened 2021-05-03
Different Black formatting results on Windows


akaihola:
I've encountered this now in a couple of pull requests: on Windows runners, Darker refuses to accept some empty lines inside functions in its own code base when running the test suite.

The two attached build logs illustrate this – look for "`FAILURES`" in the `win-py39.txt` log:
```diff
--- src/darker/tests/test_import_sorting.py
+++ src/darker/tests/test_import_sorting.py
@@ -22,7 +22,6 @@
 
             # Import when `isort` has been removed temporarily
             reload(darker.import_sorting)
-
     finally:
         # Re-import after restoring `isort` so other tests won't be affected
         reload(darker.import_sorting)
```

The same empty line is accepted fine by Ubuntu and macOS runners.

Apart from `atomicwrites` and `colorama` which are only being installed on Windows, all the same Python dependencies are installed in the test run on both Windows and Ubuntu:
```
Successfully installed appdirs-1.4.4 astroid-2.5.6 atomicwrites-1.4.0 attrs-20.3.0 black-21.4b2 click-7.1.2 colorama-0.4.4 darker-1.3.0.dev0 filelock-3.0.12 flake8-3.9.1 iniconfig-1.1.1 isort-5.8.0 lazy-object-proxy-1.6.0 mccabe-0.6.1 mypy-0.812 mypy-extensions-0.4.3 packaging-20.9 pathspec-0.8.1 pluggy-0.13.1 py-1.10.0 pycodestyle-2.7.0 pyflakes-2.3.1 pygments-2.8.1 pylint-2.8.2 pyparsing-2.4.7 pytest-6.2.3 pytest-darker-0.1.2 pytest-flake8-1.0.7 pytest-isort-2.0.0 pytest-kwparametrize-0.0.3 pytest-mypy-0.8.1 regex-2021.4.4 toml-0.10.2 typed-ast-1.4.3 typing-extensions-3.10.0.0 wrapt-1.12.1
```
[ubuntu-py39.txt](https://github.com/akaihola/darker/files/6417332/ubuntu-py39.txt)
[win-py39.txt](https://github.com/akaihola/darker/files/6417333/win-py39.txt)

akaihola:
Ping @martinRenou, here's a strange unix/windows discrepancy. I'm left clueless on how to approach this. Maybe try to reproduce on just Black?

martinRenou:
I am a Linux user myself, so I am not sure I would be of help here :S

[...]

PR #141, opened 2021-05-22
Use pytest-kwparametrize

akaihola:
Build failure on Windows is due to #133. Rebase this after merging #142.

### DIFF
--- a/src/darker/tests/test_main.py
+++ b/src/darker/tests/test_main.py
@@ -40,7 +40,7 @@ def run_isort(git_repo, monkeypatch, caplog, request):
     find_project_root.cache_clear()
 
     monkeypatch.chdir(git_repo.root)
-    paths = git_repo.add({'test1.py': 'original'}, commit='Initial commit')
+    paths = git_repo.add({"test1.py": "original"}, commit="Initial commit")
     paths["test1.py"].write_bytes(b"changed")
     args = getattr(request, "param", ())
     with patch.multiple(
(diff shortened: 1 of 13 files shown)


samoylovfp (approved):

src/darker/tests/test_main.py
```diff
@@ -76,58 +76,59 @@ def test_isort_option_with_isort_calls_sortimports(tmpdir, run_isort, isort_args
...
+    ),
     ids=["black", "black_isort", "black_unnormalize"],
 )
 @pytest.mark.parametrize("newline", ["\n", "\r\n"], ids=["unix", "windows"])
 def test_format_edited_parts(git_repo, enable_isort, black_args, newline, expect):
     """Correct reformatting and import sorting changes are produced"""
     paths = git_repo.add({"a.py": newline, "b.py": newline}, commit="Initial commit")
     paths["a.py"].write_bytes(newline.join(A_PY).encode("ascii"))
-    paths["b.py"].write_bytes("print(42 ){newline}".encode("ascii"))
+    paths["b.py"].write_bytes(f"print(42 ){newline}".encode("ascii"))
```
sherbie, line 131: Was this working before because there isn't an assertion for the newline character?

akaihola: Even now if you remove the `f` from the f-string, the test still passes. The reason is that in the `format_edited_parts()` call, only `a.py` is submitted for reformatting.

So this test also ensures that files not mentioned in `format_edited_parts(..., <changed_files>, ...)` are ignored. And in that sense, it doesn't make any difference what we write into the modified `b.py` as long as it's some Python code that *would* require reformatting if this mechanism worked wrong and didn't ignore the file.

akaihola: Even now if you remove the `f` from the f-string, the test still passes. The reason is that in the `format_edited_parts()` call, only `a.py` is submitted for reformatting.

So this test also ensures that files not mentioned in `format_edited_parts(..., <changed_files>, ...)` are ignored. And in that sense, it doesn't make any difference what we write into the modified `b.py` as long as it's some Python code that *would* require reformatting if this mechanism worked wrong and didn't ignore the file.

PR #158, opened 2021-07-09
fix: always give git_get_content_at_revision a relative path

muggenhor:
Instead of passing an absolute path to `git show COMMITISH:{path}` and being surprised that git isn't able to deal with absolute paths there.

This triggers when using `--revision <COMMIT_A>..<COMMIT_B>` where `COMMIT_B` is _not_ `:WORKTREE:`. That is what we want to use in our CI system to validate every single commit in a PR.

For C++ we've already been using this for years with `clang-format` for the purpose of ensuring that `git blame` always returns the commit that changed the logic of the code, not some follow up commit that reformatted it (by basically forbidding such follow up commits).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,6 +9,8 @@ Added
 
 Fixed
 -----
+- Ensure a full revision range ``--revision <COMMIT_A>..<COMMIT_B>`` where
+  COMMIT_B is *not* ``:WORKTREE:`` works too.
 
 
 1.2.4_ - 2021-06-27
--- a/src/darker/__main__.py
+++ b/src/darker/__main__.py
@@ -51,7 +51,9 @@ def format_edited_parts(
 
     for path_in_repo in sorted(changed_files):
         src = git_root / path_in_repo
-        rev2_content = git_get_content_at_revision(src, revrange.rev2, git_root)
+        rev2_content = git_get_content_at_revision(
+            path_in_repo, revrange.rev2, git_root
+        )
 
         # 1. run isort
         if enable_isort:
--- a/src/darker/git.py
+++ b/src/darker/git.py
@@ -40,6 +40,10 @@ def git_get_content_at_revision(path: Path, revision: str, cwd: Path) -> TextDoc
     :param cwd: The root of the Git repository
 
     """
+    assert (
+        not path.is_absolute()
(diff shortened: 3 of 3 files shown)


akaihola:
@muggenhor thanks for spotting this and submitting a fix! I'll make sure to include this in the next release.

Before merging, we still need to
- fix code formatting (see the build failures for details)
- add a change log entry
- add a test case for `format_edited_parts()` which ensures it calls `git_get_content_at_revision()` with relative paths
- (maybe even add a test case for `git_get_content_at_revision()` which shows that Git fails with absolute paths and succeeds with relative ones – this would give us a heads-up if future Git versions change in this regard)

Do you still have time to work on one or more of these?

muggenhor:
> Before merging, we still need to
>     * fix code formatting (see the build failures for details)
>     * add a change log entry

Done.

>     * add a test case for `format_edited_parts()` which ensures it calls `git_get_content_at_revision()` with relative paths
>     * (maybe even add a test case for `git_get_content_at_revision()` which shows that Git fails with absolute paths and succeeds with relative ones – this would give us a heads-up if future Git versions change in this regard)

I think both of these could be covered by an assert inside `git_get_content_at_revision` because it isn't just Git that assumes relative paths are given, that function itself also does. It even documents receiving relative paths as a precondition:
```
:param path: The relative path of the file in the Git repository
```

When I add the `assert not path.is_absolute()` statement to the front of that function it causes 77 test cases to start failing. All of those pass again with my fix added on top. So I suggest that single assert statement to function as the test case.

akaihola (approved):

akaihola:
Merged! Thanks @muggenhor for the excellent contribution.

Would you like to be invited as a collaborator to the Darker GitHub repository, and potentially help with reviewing PRs?

muggenhor:
> Merged! Thanks @muggenhor for the excellent contribution.

You're very welcome.

> Would you like to be invited as a collaborator to the Darker GitHub repository, and potentially help with reviewing PRs?

I haven't got enough time for reviewing the PRs on the projects I'm part of already. So I'm afraid that I'm reluctant to commit to more of such work.

akaihola:
@muggenhor indeed it's very wise to choose one's commitments. Anyway, all the best, and do drop a note or create an issue if Darker causes any friction!

PR #159, opened 2021-07-11
Hide expected Git errors from stderr

### LINKED ISSUE #85: Handling untracked files
Using `darker` on a repo with as of yet untracked files raises eg.:
`fatal: path 'new.py' exists on disk, but not in 'HEAD'`

Steps to reproduce
```
$ mkdir /tmp/test
$ cd /tmp/test
$ git init
$ echo "print('legacy!')" > legacy.py
$ git add .
$ git commit -m 'initial commit'

$ echo "print('new stuff!')" > new.py
$ darker .
fatal: path 'new.py' exists on disk, but not in 'HEAD'
```

I guess this might be known and intended behavior, but it would be super useful if in a legacy codebase we could just use the same formatting/linting command also for new files added.

Are there any plans to support this usecase?

akaihola:
Fixes #85.

With e.g. `darker --revision=master...`, newly created files would cause this Git error to appear on stderr:
```
fatal: Path 'newfile.py' exists on disk, but not in 'master'.
```

We already caught that situation and interpreted it properly as caused by a newly created file. In addition, the Git error messages are now silenced, but other Git errors are still logged with the `ERROR` loglevel.

Unfortunately I couldn't find a way to create a failing test, i.e. one that would be able to catch Git's stderr.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,8 @@ Fixed
 -----
 - Ensure a full revision range ``--revision <COMMIT_A>..<COMMIT_B>`` where
   COMMIT_B is *not* ``:WORKTREE:`` works too.
+- Hide fatal error from Git on stderr when ``git show`` doesn't find the file in rev1.
+  This isn't fatal from Darker's point of view since it's a newly created file.
 
 
 1.2.4_ - 2021-06-27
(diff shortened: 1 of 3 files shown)


akaihola:
@casio would you be interested to review this PR? I invited you as collaborator in this project in case you would.

akaihola:
@flying-sheep, that's an excellent catch! Could you please open a separate issue about that?

Issue #102, opened 2020-12-01
No error if paths on command line match no dirs/files


akaihola:
```bash
$ ls
src setup.py
$ darker foobar
```

samoylovfp:
I wonder if there is a case when we should NOT error out

PR #163, opened 2021-07-14
Validate that paths given on command line do exist

### LINKED ISSUE #147: Darker running with --diff and --quiet on VSCode
I've set up darker on VSCode the way it says on the tutorial but even if I don't pass `--diff` the command stills run with `--diff` and `--quiet`. I've seen there's a [PR](https://github.com/akaihola/darker/pull/84) open for that but I also would like to know if it's possible to run the command without `--diff` and `--quiet` explicitly. Thanks

akaihola:
~Disallow paths starting with a hyphen – those are probably due to quoting multiple arguments like `"--lint pylint"`.~
Disallow paths which don't exist as files or directories on disk (if rev2 is `WORKTREE`) or in a past revision (if rev2 points to a commit). Fixes #147 and #102.

- [x] unit tests for `darker.git.get_missing_at_revision()` for commits
- [x] solve Windows test failure for `darker.git.get_missing_at_revision()` unit test
- [x] unit tests for `darker.git.get_missing_at_revision()` for `WORKTREE`
- [x] unit tests for `darker.git._git_exists_in_revision()`
- [x] unit tests for `darker.__main__.main()` failing for non-existing files in a revision or the working tree
- [x] change log

### DIFF
--- a/src/darker/command_line.py
+++ b/src/darker/command_line.py
@@ -1,4 +1,6 @@
-from argparse import ArgumentParser, Namespace
+"""Command line parsing for the ``darker`` binary"""
+
+from argparse import Action, ArgumentError, ArgumentParser, Namespace
 from typing import Any, List, Optional, Text, Tuple
 
 from darker import help as hlp
@@ -105,6 +107,14 @@ def parse_command_line(argv: List[str]) -> Tuple[Namespace, DarkerConfig, Darker
     #    This is used to find out differences between the effective configuration and
     #    default configuration values, and print them out in verbose mode.
     parser_with_original_defaults = make_argument_parser(require_src=True)
+    invalid_src_paths = [src for src in args.src if src.startswith("-")]
+    if invalid_src_paths:
+        invalid_path_reprs = " ".join(repr(path) for path in invalid_src_paths)
+        raise ArgumentError(
+            Action(["src"], dest="src", metavar="PATH"),
+            f"PATH can't begin with a hyphen: {invalid_path_reprs}\n"
+            "Maybe check your shell quoting?",
+        )
     return (
         args,
         get_effective_config(args),
--- a/src/darker/command_line.py
+++ b/src/darker/command_line.py
@@ -1,6 +1,6 @@
 """Command line parsing for the ``darker`` binary"""
 
-from argparse import Action, ArgumentError, ArgumentParser, Namespace
+from argparse import ArgumentParser, Namespace
 from typing import Any, List, Optional, Text, Tuple
 
 from darker import help as hlp
@@ -107,14 +107,6 @@ def parse_command_line(argv: List[str]) -> Tuple[Namespace, DarkerConfig, Darker
     #    This is used to find out differences between the effective configuration and
     #    default configuration values, and print them out in verbose mode.
     parser_with_original_defaults = make_argument_parser(require_src=True)
-    invalid_src_paths = [src for src in args.src if src.startswith("-")]
-    if invalid_src_paths:
-        invalid_path_reprs = " ".join(repr(path) for path in invalid_src_paths)
-        raise ArgumentError(
-            Action(["src"], dest="src", metavar="PATH"),
-            f"PATH can't begin with a hyphen: {invalid_path_reprs}\n"
-            "Maybe check your shell quoting?",
-        )
     return (
         args,
         get_effective_config(args),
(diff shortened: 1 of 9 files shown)


akaihola:
@hlrossato, care to take a look at this pull request and check that it makes sense?

src/darker/command_line.py
```diff
@@ -105,6 +107,14 @@ def parse_command_line(argv: List[str]) -> Tuple[Namespace, DarkerConfig, Darker
     #    This is used to find out differences between the effective configuration and
     #    default configuration values, and print them out in verbose mode.
     parser_with_original_defaults = make_argument_parser(require_src=True)
+    invalid_src_paths = [src for src in args.src if src.startswith("-")]
```
levouh, line 110: Can you just check something like `os.path.exists(...)`? This seems like it would cover more use cases than just handling them one-by-one as things get messed up as is referenced in #147

akaihola: Indeed! This would also be more in line with what Black does. For non-existing files on the command line it terminates with
```
$ black nonexisting.py
Usage: black [OPTIONS] SRC ...
Try 'black -h' for help.

Error: Invalid value for 'SRC ...': Path 'nonexisting.py' does not exist.
```
and for missing directories
```
$ black nonexisting/
Usage: black [OPTIONS] SRC ...
Try 'black -h' for help.

Error: Invalid value for 'SRC ...': Path 'nonexisting/' does not exist.
```
I'll modify the patch to do the same in Darker as well.

akaihola: ...although in Darker's case it's a bit more nuanced than that. Let's say you do
```
$ darker --revision master..HEAD oldfile.py
```
but you've deleted `oldfile.py` in your working tree. Darker should still check the file contents between `master` and the last commit (`HEAD`).

This is still doable, just requires a bit more thinking on where to check for existence of files and directories.

akaihola: In [darker.git.git_get_modified_files()](https://github.com/akaihola/darker/blob/508a6ec169d9add0e6c1123f0f20f7fc6829a3f1/src/darker/git.py#L169-L175), Darker calls
```
git diff --name-only --relative <rev> -- <path(s)>
```
for changed files, and additionally
```
git ls-files --others --exclude-standard -- <path(s)>
```
if comparing against the working tree so newly created Python files not yet tracked with Git are also processed.

Unfortunately, `git diff` gives empty output both for unchanged and non-existing file paths. So we'd need to introduce a third Git invocation to check which files are missing from `rev2` (if `rev2` is not `WORKTREE`).

In case `rev2` is `WORKTREE`, a simple filesystem existence check does the trick.

akaihola: @levouh I implemented the existence check both for the working tree (filesystem) and arbitrary Git commits. How does it look to you now?

akaihola: @levouh, what about this scenario:

- `file.py` exists in `commit1`
- `file.py` is removed in `commit2`
- user does `darker --diff --revision commit1..commit2`

Should this exit with an error or succeed silently? Surely there's nothing to report since there are no "modified lines" left in a non-existing file.

With the new `--stdout` option (#165), this surely shouldn't succeed with empty output, since that would indicate that an empty `file.py` *does* exist in `commit2`.

akaihola:
In git-for-windows/git#3344, I reported the fact that `git cat-file` doesn't understand backslash as path separator. We needed to work around that by using `path.as_posix()` instead of `str(path)` in [darker.git._git_exists_in_revision()](https://github.com/akaihola/darker/blob/1edd63888dc19d70a07d58e724b2b0dd55a997dd/src/darker/git.py#L169-L182):
```python
    # Surprise: On Windows, `git cat-file` doesn't work with backslash directory
    # separators in paths. We need to use Posix paths and forward slashes instead.
    cmd = ["git", "cat-file", "-e", f"{rev2}:{path.as_posix()}"]
    result = run(cmd, check=False, stderr=DEVNULL)
```

I noticed that `git show` has the same behavior, so we need a Windows compatibility fix for that one as well.

At the same time, `git diff`, `git ls-files` and `git log` work just fine with backslashes.

src/darker/command_line.py
akaihola: @levouh, FYI, I decided to go with the design in which existence in `rev2` is validated when `--revision=rev1..rev2` is specified.

PR #165, opened 2021-07-17
Add the `-d` / `--stdout` option. Fixes #164.

### LINKED ISSUE #164: New file to stdout
Is it possible to pipe the new file to stdout? Would be nice so that one could use filter in vim :)

akaihola:
- [x] decide on the design (see [this comment](https://github.com/akaihola/darker/issues/164#issuecomment-881945048))
- [x] only print out a single file
- [x] unit tests
- [x] change log

### DIFF
--- a/src/darker/__main__.py
+++ b/src/darker/__main__.py
@@ -190,6 +190,7 @@ def print_source(new: TextDocument) -> None:
     """Print the reformatted Python source code"""
     if sys.stdout.isatty():
         try:
+            # pylint: disable=import-outside-toplevel
             from pygments import highlight
             from pygments.formatters import TerminalFormatter
             from pygments.lexers.python import PythonLexer
--- a/src/darker/__main__.py
+++ b/src/darker/__main__.py
@@ -35,6 +35,7 @@ def format_edited_parts(
     revrange: RevisionRange,
     enable_isort: bool,
     black_args: BlackArgs,
+    report_unmodified: bool,
 ) -> Generator[Tuple[Path, TextDocument, TextDocument], None, None]:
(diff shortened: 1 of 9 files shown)


src/darker/__main__.py
```diff
@@ -233,7 +249,7 @@ def main(argv: List[str] = None) -> int:
     failures_on_modified_lines = False
 
     revrange = RevisionRange.parse(args.revision)
-    write_modified_files = not args.check and not args.diff
+    write_modified_files = not args.check and not args.diff and not args.stdout
```
levouh, line 252: Somewhere around here, should there be a check that `len(paths) == 1` when `args.stdout` is truthy?

src/darker/config.py
```diff
@@ -31,21 +32,33 @@ class OutputMode:
...
             return cls.DIFF
         if args.stdout:
             return cls.CONTENT
         return cls.NOTHING
 
     @staticmethod
-    def validate(diff: bool, stdout: bool) -> None:
-        """Raise an exception if both ``diff`` and ``stdout`` are enabled"""
+    def validate_diff_stdout(diff: bool, stdout: bool) -> None:
+        """Raise an exception if ``diff`` and ``stdout`` options are bothenabled"""
```
levouh, line 58: _nit_: `bothenabled` -> `both enabled`

levouh (commented):
Useless nitpick @akaihola, going to pull/test and then I'll report back here. Otherwise LGTM.

src/darker/config.py
akaihola: Thanks, fixed.

akaihola:
@levouh, I rebased on `master` after merging another branch.

PR #171, opened 2021-07-30
Respect black exclude

### LINKED ISSUE #146: Not ignoring files even though it is excluded in Black config
So here's a sample config for Black and `setup.py` is ignored by Black when you run it.

```toml
[tool.black]
extend-exclude = '''
# A regex preceded with ^/ will apply only to files and directories
# in the root of the project.
^/setup.py
```

However, darker is not respecting this configuration and still reformats `setup.py`

Additional detail: darker is set in Pycharm's File Watcher so it keeps reformatting excluded files.

akaihola:
Skip Black reformatting for files matching the `exclude`, `extend-exclude` or `force-exclude` Black configuration options. Fixes #146.

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -75,3 +75,6 @@ ignore_missing_imports = True
 
 [mypy-pygments.lexers.*]
 ignore_missing_imports = True
+
+[mypy-regex]
+ignore_missing_imports = True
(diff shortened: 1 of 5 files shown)


roniemartinez:
Looks good. 🥇

PR #172, opened 2021-08-02
Use forward slash as the path separator for Git on Windows

akaihola:
This was found and worked around in #163.

Git behavior reported in git-for-windows/git/issues/3344.


akaihola:
The `macos-latest, 3.8` [build failure today](https://github.com/akaihola/darker/pull/172/checks?check_run_id=3218212179) was due to an insufficient delay threshold when asserting file timestamps in Git related tests. This will be fixed later in #173.

Issue #26, opened 2020-07-11
Profile performance issues


akaihola:
In #23, @DylanYoung mentions:
> @Mystic-Mirage Do you use darker in your editor? It takes like 30 seconds to format.

@Mystic-Mirage [responds](https://github.com/akaihola/darker/issues/23#issuecomment-656730731):
I'm a lucky guy and my darker takes 0.5~1.5 seconds. My IDE calls it on every save.

We need to profile darker and find bottlenecks.

akaihola:
@DylanYoung,

Are you still using Darker in your editor? Have you upgraded all the way up to Darker 1.2.1? Do you still have the performance problems you mentioned in #23?

akaihola:
@DylanYoung, we have some performance improvements coming up. You could install Darker from the [rogalski/darker@jobs](https://github.com/rogalski/darker/tree/jobs) branch and use the `--jobs=0` argument (zero causes it to use one job per core) and see if that makes it faster for you.

In any case I think it's still a good idea for us to do profiling for some real-world use cases. I'm keeping this open and assigning to a future milestone.

Issue #23, opened 2020-07-09
Support for blackd


DylanYoung:
Probably this requires a port argument and not much else after #22 is finished. 

Then the arguments just need to be translated into an HTTP request with the appropriate X- headers.

Thoughts?

akaihola:
Hi @DylanYoung,

Thanks for the idea!

According to [documentation](https://black.readthedocs.io/en/stable/blackd.html), the main benefit from using `blackd` instead of black to get rid of the startup overhead and have a faster response in IDEs.

`darker` currently imports `black` and calls it inside the Python process instead of invoking `black` as a shell command. Thus we only spend the Python startup overhead once.

If `darker` supported `blackd`, you'd still pay the startup overhead of `darker` and only get a small speedup from not having to import `black`.

To achieve similar fast response as with `blackd`, we would have to implement a `darkerd` and do API calls to `blackd` from it.

Or am I missing a separate non-performance-related benefit from using `blackd`?

akaihola:
On the topic of a potential `darkerd` implementation:

Let's consider a `darkerd` daemon which makes HTTP API calls to a separate `blackd` daemon. That would mean
- two processes running,
- added code complexity in `darkerd`, and
- probably no performance benefits (compared to calling the `black` Python API directly in the `darkerd` process).

So the performance benefit would be achieved by just
- creating a `darkerd`,
- making it call the `darker` Python API, and
- letting `darker` in turn call the `black` and (and optionally `isort`) Python APIs internally.

Compared to `blackd`, it's important to consider that it's not possible to pass only source code in the HTTP request. `darker` needs to have access to the unmodified version of each file in Git `HEAD`. Do do this,
- the HTTP request  to `darkerd` must point to actual files on disk (either Git root dir + set of relative paths, or just a set of absolute paths)
- the `darkerd` process must have access to the filesystem, so in practice run in the same machine as the caller (i.e. the IDE)

If we were to implement `darkerd`, I would
- mirror the implementation of `blackd` (it's under 200 lines of code)
- use the same libraries (`aiohttp`, `aiohttp-cors` and maybe `click`)

How does this reasoning sound?

Mystic-Mirage:
A lot of code without a valued benefit.
Does somebody really face performance issues?

DylanYoung:
Do you use darker in your editor?  It takes like 30 seconds to format.

DylanYoung:
[...]

PR #179, opened 2021-08-18
--jobs argument

### LINKED ISSUE #178: Missing `--jobs` support
Monitoring the resource consumption with multiple Python files staged in a pre-commit scenario, the CPU is not used very effectively, since only a single subprocess is invoked. There should be a way to parallelize the workload over separate files, utilizing all available CPU cores to achieve better performance.

rogalski:
TBD, closes #178 

- [x] tests
- [x] change log
- [x] documentation
- [x] measure peformance with one job, many jobs using threads, and many jobs using processes
- [x] decide whether to use threads or processes -> **processes**
- [x] use `ProcessPoolExecutor`
- [x] use synchronous calls when `--jobs=1`
- [x] `--jobs=1` as the default
- [x] rebase on `master` to resolve new merge conflicts

### DIFF
--- a/src/darker/__main__.py
+++ b/src/darker/__main__.py
@@ -8,6 +8,7 @@ from datetime import datetime
 from difflib import unified_diff
 from pathlib import Path
 from typing import Collection, Generator, List, Tuple
+import concurrent.futures
 
 from darker.black_diff import (
     BlackConfig,
@@ -40,6 +41,8 @@ from darker.verification import ASTVerifier, BinarySearch, NotEquivalentError
 
 logger = logging.getLogger(__name__)
 
+ProcessedDocument = Tuple[Path, TextDocument, TextDocument]
+
 
 def format_edited_parts(
     root: Path,
@@ -49,7 +52,8 @@ def format_edited_parts(
     enable_isort: bool,
     black_config: BlackConfig,
     report_unmodified: bool,
-) -> Generator[Tuple[Path, TextDocument, TextDocument], None, None]:
+    jobs: int = 1,
+) -> Generator[ProcessedDocument, None, None]:
     """Black (and optional isort) formatting for chunks with edits since the last commit
 
     Files inside given directories and excluded by Black's configuration are not
@@ -67,42 +71,70 @@ def format_edited_parts(
     :param enable_isort: ``True`` to also run ``isort`` first on each changed file
     :param black_config: Configuration to use for running Black
     :param report_unmodified: ``True`` to yield also files which weren't modified
+    :param jobs: number of cpu processes to use (0 - autodetect)
     :return: A generator which yields details about changes for each file which should
              be reformatted, and skips unchanged files.
 
     """
-    for path_in_repo in sorted(changed_files):
-        src = root / path_in_repo
-        rev2_content = git_get_content_at_revision(path_in_repo, revrange.rev2, root)
-
-        # 1. run isort
-        if enable_isort:
-            rev2_isorted = apply_isort(
-                rev2_content,
-                src,
-                black_config.get("config"),
-                black_config.get("line_length"),
-            )
-        else:
-            rev2_isorted = rev2_content
-        if path_in_repo not in black_exclude:
-            # 9. A re-formatted Python file which produces an identical AST was
-            #    created successfully - write an updated file or print the diff if
-            #    there were any changes to the original
-            content_after_reformatting = _reformat_single_file(
+    with concurrent.futures.ThreadPoolExecutor(max_workers=jobs or None) as executor:
+        futures: List[
+            concurrent.futures.Future[ProcessedDocument]
+        ] = []  # pylint: disable=unsubscriptable-object
+        for path_in_repo in sorted(changed_files):
+            future = executor.submit(
+                _format_single_file,
                 root,
                 path_in_repo,
+                black_exclude,
                 revrange,
-                rev2_content,
-                rev2_isorted,
-                enable_isort,
                 black_config,
+                enable_isort,
             )
-        else:
-            # File was excluded by Black configuration, don't reformat
-            content_after_reformatting = rev2_isorted
-        if report_unmodified or content_after_reformatting != rev2_content:
-            yield (src, rev2_content, content_after_reformatting)
+            futures.append(future)
+
+        for future in concurrent.futures.as_completed(futures):
+            src, rev2_content, content_after_reformatting = future.result()
+            if report_unmodified or content_after_reformatting != rev2_content:
+                yield (src, rev2_content, content_after_reformatting)
+
+
+def _format_single_file(
+    git_root: Path,
+    path_in_repo: Path,
+    black_exclude: Collection[Path],  # pylint: disable=unsubscriptable-object
+    revrange: RevisionRange,
+    black_config: BlackConfig,
+    enable_isort: bool,
+) -> ProcessedDocument:
+    src = git_root / path_in_repo
+    rev2_content = git_get_content_at_revision(path_in_repo, revrange.rev2, git_root)
+    # 1. run isort
+    if enable_isort:
+        rev2_isorted = apply_isort(
+            rev2_content,
+            src,
+            black_config.get("config"),
+            black_config.get("line_length"),
+        )
+    else:
+        rev2_isorted = rev2_content
+    if path_in_repo not in black_exclude:
+        # 9. A re-formatted Python file which produces an identical AST was
+        #    created successfully - write an updated file or print the diff if
+        #    there were any changes to the original
+        content_after_reformatting = _reformat_single_file(
+            git_root,
+            path_in_repo,
+            revrange,
+            rev2_content,
+            rev2_isorted,
+            enable_isort,
+            black_config,
+        )
+    else:
+        # File was excluded by Black configuration, don't reformat
+        content_after_reformatting = rev2_isorted
+    return src, rev2_content, content_after_reformatting
 
 
 def _reformat_single_file(  # pylint: disable=too-many-arguments,too-many-locals
@@ -391,6 +423,7 @@ def main(argv: List[str] = None) -> int:
         args.isort,
         black_config,
         report_unmodified=output_mode == OutputMode.CONTENT,
+        jobs=config["jobs"],
     ):
         formatting_failures_on_modified_lines = True
         if output_mode == OutputMode.DIFF:
--- a/src/darker/__main__.py
+++ b/src/darker/__main__.py
@@ -18,6 +18,7 @@ from darker.black_diff import (
 )
 from darker.chooser import choose_lines
 from darker.command_line import parse_command_line
+from darker.concurrency import get_executor
 from darker.config import OutputMode, dump_config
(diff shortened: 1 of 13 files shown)


rogalski:
@overratedpro FYI

overratedpro (approved):
I like this very much, great reduction in execution times! :+1:

src/darker/__main__.py
```diff
@@ -56,45 +60,70 @@ def format_edited_parts(
...
+            src, rev2_content, content_after_reformatting = future.result()
+            if report_unmodified or content_after_reformatting != rev2_content:
+                yield (src, rev2_content, content_after_reformatting)
+
+
+def _format_single_file(
+    git_root: Path,
+    path_in_repo: Path,
+    revrange: RevisionRange,
+    black_config: BlackConfig,
```
akaihola, line 95: Since `format_edited_parts` can pass `black_config==None`, I believe this should be: 
```suggestion
    black_config: Optional[BlackConfig],
```

src/darker/__main__.py
```diff
@@ -56,45 +60,70 @@ def format_edited_parts(
...
+            enable_isort,
+            black_config,
+        )
+    else:
+        # File was excluded by Black configuration, don't reformat
+        content_after_reformatting = rev2_isorted
+    return src, rev2_content, content_after_reformatting
 
 
 def _reformat_single_file(  # pylint: disable=too-many-arguments,too-many-locals
```
akaihola, line 129: It would make sense to rename this function now to `_apply_black()` since `_format_single_file()` was now introduced and we also already have `apply_isort()`.

akaihola:
Hi @rogalski,

Thanks for the pull request, this is a great enhancement!

The test failure is due to an incompatibility with Black 21.8b0. I've fixed it in #183. Could you review, and rebase this on `master` after we've merged the fix? I invited you as a collaborator on this repository.

src/darker/__main__.py
```diff
@@ -7,6 +7,7 @@
 from difflib import unified_diff
 from pathlib import Path
 from typing import Collection, Generator, List, Tuple
+import concurrent.futures
```
akaihola, line 10: isort wants to move this before `import logging`.

src/darker/__main__.py
```diff
@@ -56,45 +60,70 @@ def format_edited_parts(
...
+) -> ProcessedDocument:
+    src = git_root / path_in_repo
+    rev2_content = git_get_content_at_revision(path_in_repo, revrange.rev2, git_root)
+    # 1. run isort
+    if enable_isort:
+        rev2_isorted = apply_isort(
+            rev2_content,
+            src,
+            black_config.get("config"),
+            black_config.get("line_length"),
```
akaihola, line 106: `black.config.get()` will fail if `black_config is None`.

src/darker/__main__.py
akaihola: See below, probably not after all. Instead, add a separate `enable_black: bool` argument.

akaihola:
For some reason the order in which files appear in `darker --diff` output changes with this patch. This fix is required for tests:
```diff
diff --git a/src/darker/tests/test_main_revision.py b/src/darker/tests/test_main_revision.py
index 6374551..4472236 100644
--- a/src/darker/tests/test_main_revision.py
+++ b/src/darker/tests/test_main_revision.py
@@ -43,63 +43,63 @@ from darker.tests.helpers import raises_if_exception
     dict(
         revision="",
         worktree_content=b"USERMOD=1\n",
-        expect=["+1", "+1M0", "+2-1", "+2", "+2M1-0", "+2M1"],
+        expect={"+1", "+1M0", "+2-1", "+2", "+2M1-0", "+2M1"},
     ),
     dict(
         revision="",
         worktree_content=b"ORIGINAL=1\n",
-        expect=["+1M0", "+2-1", "+2M1-0", "+2M1"],
+        expect={"+1M0", "+2-1", "+2M1-0", "+2M1"},
     ),
     dict(
         revision="",
         worktree_content=b"MODIFIED=1\n",
-        expect=["+1", "+2-1", "+2", "+2M1-0"],
+        expect={"+1", "+2-1", "+2", "+2M1-0"},
     ),
     dict(
         revision="HEAD",
         worktree_content=b"USERMOD=1\n",
-        expect=["+1", "+1M0", "+2-1", "+2", "+2M1-0", "+2M1"],
+        expect={"+1", "+1M0", "+2-1", "+2", "+2M1-0", "+2M1"},
     ),
     dict(
         revision="HEAD",
         worktree_content=b"ORIGINAL=1\n",
-        expect=["+1M0", "+2-1", "+2M1-0", "+2M1"],
+        expect={"+1M0", "+2-1", "+2M1-0", "+2M1"},
     ),
     dict(
         revision="HEAD",
         worktree_content=b"MODIFIED=1\n",
-        expect=["+1", "+2-1", "+2", "+2M1-0"],
+        expect={"+1", "+2-1", "+2", "+2M1-0"},
     ),
     dict(
         revision="HEAD^",
         worktree_content=b"USERMOD=1\n",
-        expect=["+1", "+1M0", "+2-1", "+2", "+2M1-0", "+2M1"],
+        expect={"+1", "+1M0", "+2-1", "+2", "+2M1-0", "+2M1"},
     ),
     dict(
         revision="HEAD^",
         worktree_content=b"USERMOD=1\n",
-        expect=["+1", "+1M0", "+2-1", "+2", "+2M1-0", "+2M1"],
+        expect={"+1", "+1M0", "+2-1", "+2", "+2M1-0", "+2M1"},
     ),
     dict(
         revision="HEAD^",
         worktree_content=b"ORIGINAL=1\n",
-        expect=["+2-1", "+2M1-0", "+2M1"],
+        expect={"+2-1", "+2M1-0", "+2M1"},
     ),
     dict(
         revision="HEAD^",
         worktree_content=b"MODIFIED=1\n",
-        expect=["+1", "+1M0", "+2-1", "+2"],
+        expect={"+1", "+1M0", "+2-1", "+2"},
     ),
     dict(
         revision="HEAD~2",
         worktree_content=b"USERMOD=1\n",
-        expect=["+1", "+1M0", "+2-1", "+2", "+2M1-0", "+2M1"],
+        expect={"+1", "+1M0", "+2-1", "+2", "+2M1-0", "+2M1"},
     ),
-    dict(revision="HEAD~2", worktree_content=b"ORIGINAL=1\n", expect=["+1", "+1M0"]),
+    dict(revision="HEAD~2", worktree_content=b"ORIGINAL=1\n", expect={"+1", "+1M0"}),
     dict(
         revision="HEAD~2",
         worktree_content=b"MODIFIED=1\n",
-        expect=["+1", "+1M0", "+2-1", "+2", "+2M1-0", "+2M1"],
+        expect={"+1", "+1M0", "+2-1", "+2", "+2M1-0", "+2M1"},
     ),
     dict(revision="HEAD~3", worktree_content=b"USERMOD=1\n", expect=SystemExit),
 )
@@ -141,9 +141,9 @@ def test_revision(git_repo, monkeypatch, capsys, revision, worktree_content, exp
     with raises_if_exception(expect):
         main(arguments)
 
-        modified_files = [
+        modified_files = {
             line.split("\t")[0][4:-3]
             for line in capsys.readouterr().out.splitlines()
             if line.startswith("+++ ")
-        ]
+        }
         assert modified_files == expect
```

rogalski:
@akaihola Let's postpone until we merge all of actual core performance improvements.
We need set of data (baseline + patch) which triggers pessimistic performance and/or long bisect first.

akaihola:
@rogalski, I found and fixed one more issue: with `--diff --jobs N ` (N>1), the diff output wouldn't be in sorted order anymore. This would cause tests to sometimes fail (see e.g. [here](https://github.com/akaihola/darker/runs/4050585415?check_suite_focus=true)), depending on timing.

Also, the `--jobs` option isn't yet documented in `README.rst` – I'll add that as well.

akaihola:
@rogalski, I'm getting bad results with this branch. I ran Darker on its own codebase with `--diff --isort` 5 times with the following setups:
- `master` branch
- `jobs` branch with `--jobs 1` and `ThreadPoolExecutor`
- `jobs` branch with `--jobs 4` and `ThreadPoolExecutor`
- `jobs` branch with `--jobs 8` and `ThreadPoolExecutor`
- `jobs` branch with `--jobs 1` and `ProcessPoolExecutor`
- `jobs` branch with `--jobs 4` and `ProcessPoolExecutor`
- `jobs` branch with `--jobs 8` and `ProcessPoolExecutor`

The average results are:
- `--jobs 1` is about **30x slower** than `master` with both threads and processes
- `--jobs 4` is 5% slower than `--jobs 1` with threads
- `--jobs 8` is 7% slower than `--jobs 1` with threads
- `--jobs 4` is 6% faster than `--jobs 1` with processes
- `--jobs 8` is 5% faster than `--jobs 1` with processes

Based on the ~30x slowdown of this branch compared to `master`, I think we need to take a closer look on what on earth is going on. Maybe the code refactoring, not multiple jobs, changed the behavior somehow?

Maybe we need to profile (see #26) to understand the bottlenecks better.

In any case, processes seem to beat threads hands down.

akaihola:
I made a version with a dummy executor like this:
```python
class DummyExecutor(concurrent.futures.Executor):
    def __init__(self, max_workers=None):
        self._shutdown = False

    def submit(self, fn, *args, **kwargs):
        if self._shutdown:
            raise RuntimeError("cannot schedule new futures after shutdown")

        f = concurrent.futures.Future()
        try:
            result = fn(*args, **kwargs)
        except BaseException as e:
            f.set_exception(e)
        else:
            f.set_result(result)

        return f

    def shutdown(self, wait=True):
        self._shutdown = True
```

This time I ran without `--lint pylint`, and compared to `master`:
- `DummyExecutor` is **4.8x slower**
- `ThreadPoolExecutor` with `--jobs 1` is 4.6x slower
- `ProcessPoolExecutor` with `--jobs 1` is 5.7x slower

So obviously the refactoring somehow adds a 3.8x overhead. I'll review the code again and try to understand what is causing that.

akaihola:
I verified that sorting of the files I added in b5d0ffdbab612a6c85f1c238f24c802c490f5ead didn't affect performance.

akaihola:
@rogalski @overratedpro, I worked on this pull request a bit during the past couple of days. It's now again rebased on `master`, and I did some modifications and additions. I'm willing to include it in the next release (1.4.0) if you see no serious flaws in it and feel like it will be useful to you and perhaps other users.

My main changes in a nutshell:
- `--jobs=1` is the default
- `--jobs=1` runs synchronously (using a custom dummy executor)
- use processes instead of threads for jobs

akaihola:
@DylanYoung I remember you had performance issues back in #23. Some performance improvements have already been done, and this one adds the capability to process multiple files in parallel.

Would you be interested to test this approach and to review the pull request?

akaihola:
@overratedpro quite a bit has changed in this PR since your review. Would you have time to take another look?

akaihola:
I'm moving this one to the [1.5.0 release](https://github.com/akaihola/darker/milestone/13). I extracted some of the CI changes into #303 which will go into [1.4.2](https://github.com/akaihola/darker/milestone/14).

overratedpro (approved):
LGTM!

PR #183, opened 2021-08-31
Fix compatibility with Black >= 21.7b1.dev9

akaihola:
Duplicate of #189.

### DIFF
--- a/src/darker/black_diff.py
+++ b/src/darker/black_diff.py
@@ -69,6 +69,7 @@ DEFAULT_EXCLUDE_RE = re_compile_maybe_verbose(DEFAULT_EXCLUDES)
 DEFAULT_INCLUDE_RE = re_compile_maybe_verbose(DEFAULT_INCLUDES)
 BLACK_VERSION = LooseVersion(black_version)
 
+
 class BlackConfig(TypedDict, total=False):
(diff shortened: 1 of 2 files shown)


src/darker/black_diff.py
```diff
@@ -35,13 +35,15 @@
 
 import logging
 import sys
+from distutils.version import LooseVersion
```
flying-sheep, line 38: I wouldn’t use this class. its comparison operators just return whatever if a version doesn’t parse right.

These days I always use [`packaging.version`](https://packaging.pypa.io/en/latest/version.html).

I think #189 is a simpler fix here.

flying-sheep, line 38: I think #189 is a simpler fix here, but if you prefer doing it by version I wouldn’t use this class. Its comparison operators just return whatever if a version doesn’t parse right.

These days I always use [`packaging.version`](https://packaging.pypa.io/en/latest/version.html).

flying-sheep (changes requested):

src/darker/black_diff.py
akaihola: `packaging` isn't currenly required by Darker, and it feels wrong to add that dependency just because of this compatibility hack. Your approach in #189 makes more sense.

PR #187, opened 2021-09-02
Add a sidebar to README

akaihola:


akaihola:
The latest test failures will be fixed by #190.

PR #188, opened 2021-09-03
Rely on rST option list support

flying-sheep:
rST has built in support for option lists, maybe we should use it?

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -247,48 +247,46 @@ For more details, see:
 
 The following `command line arguments`_ can also be used to modify the defaults:
 
-.. code-block:: shell
+-r REVISION, --revision REVISION
+                      Git revision against which to compare the working tree.
+                      Tags, branch names, commit hashes, and other expressions
+                      like HEAD~5 work here. Also a range like master...HEAD or
+                      master... can be used to compare the best common ancestor.
+                      With the magic value :PRE-COMMIT:, Darker expects the
+                      revision range from the PRE_COMMIT_FROM_REF and
+                      PRE_COMMIT_TO_REF environment variables.
 
-     -r REVISION, --revision REVISION
-                           Git revision against which to compare the working tree.
-                           Tags, branch names, commit hashes, and other expressions
-                           like HEAD~5 work here. Also a range like master...HEAD or
-                           master... can be used to compare the best common ancestor.
-                           With the magic value :PRE-COMMIT:, Darker expects the
-                           revision range from the PRE_COMMIT_FROM_REF and
-                           PRE_COMMIT_TO_REF environment variables.
+--diff                Don't write the files back, just output a diff for
+                      each file on stdout. Highlight syntax on screen if
+                      the `pygments` package is available.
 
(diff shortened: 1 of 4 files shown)


akaihola:
Neat, I didn't know about that!

This is how it renders for me on GitHub:
image

The table seems to follow the layout of the reST source – short options are in the leftmost column with the description in the rightmost one. Longer options are on their own row, and the leftmost column is empty on the following row with the description in the rightmost one. This looks a bit incostintent to me. Is there any other way to get the options always in the leftmost column than to indent and squeeze descriptions to the right of the options in the reST source?

Also, you've kept the original single-backtick highlighting from `darker --help` (for e.g. package names). GitHub doesn't format these differently at all – we probably need to use double backticks, right?

The single drawback from using this format and customizing it is that it's no longer possible to do a straight copy-paste from `darker --help` to `README.rst` whenever options have been modified. But that's arguably an acceptable trade-off for the more pleasing look for the eye.

flying-sheep:
> Is there any other way to get the options always in the leftmost column than to indent and squeeze descriptions to the right of the options in the reST source?

It directly mirrors the source code. Long options are written as

```rst
-S, --skip-string-normalization
                      Don't normalize string quotes or prefixes
```

and short ones as

```rst
-i, --isort           Also sort imports using the `isort` package
```

so I assume indenting everything until even the longest options fit would achieve what you want.

> The single drawback from using this format and customizing it is that it's no longer possible to do a straight copy-paste from `darker --help` to `README.rst` whenever options have been modified.

Doing `ArgumentParser(..., formatter_class=lambda **kwargs: NewlinePreservingFormatter(**kwargs, max_help_position=float('inf')))` works!

akaihola:
@flying-sheep, I'm experimenting with code for generating the list of options in the format discussed above. Let me commit it to this branch so we can review if it makes sense.

akaihola:
There you go, `darker --options-for-readme` should now output the list of options in a format which can be copy-pasted as such into `README.rst`. It wasn't super straightforward to get this, it required using some `argparse` internals. But it works...

akaihola:
...however, what the hell, it still renders like this:
image

So probably GitHub's renderer makes its own calls for how to display long options!

flying-sheep:
huh, seems like it! or even docutils … too bad!

akaihola:
Yes, and also if we write options on their own line with description starting indented on the next line, GitHub and docutils will still keep the rendering same.

I guess let's live with that.

The test failures will be fixed by #190.

flying-sheep:
do you prefer squash & merge or a merge commit?

akaihola:
My workflow has been to rebase feature branches on `master` and sometimes reorganize the commits to get rid of extra back-and-forths and trials and errors. I then do a `--no-ff` merge commit to preserve the feature branch and its commits in the repository history.

But that's not a hard requirement, I don't want to be too strict on this.

akaihola (approved):
This is a nice clean-up for the README. We could even add CI automation for updating (or at least verifying) the command line usage in README. But this is a great first step!

PR #189, opened 2021-09-03
Fix gen_python_files usage for black>=21.7b1.dev9

flying-sheep:
This should fix the tests and be backwards compatible

### DIFF
--- a/src/darker/black_diff.py
+++ b/src/darker/black_diff.py
@@ -133,7 +133,7 @@ def apply_black_excludes(
     """
     sig = inspect.signature(gen_python_files)
     # those two exist and are required in black>=21.7b1.dev9
-    kwa = dict(verbose=False, quiet=False) if "verbose" in sig.parameters else {}
+    kwargs = dict(verbose=False, quiet=False) if "verbose" in sig.parameters else {}
     return set(
         gen_python_files(
             (root / path for path in paths),
@@ -144,7 +144,7 @@ def apply_black_excludes(
             force_exclude=black_config.get("force_exclude"),
             report=Report(),
             gitignore=None,
-            **kwa,
+            **kwargs,
         )
     )
 
--- a/src/darker/black_diff.py
+++ b/src/darker/black_diff.py
@@ -32,7 +32,7 @@ for how this result is further processed with:
   based on whether reformats touch user-edited lines
 
 """
-
+import inspect
 import logging
 import sys
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


akaihola:
@flying-sheep oops, I already had a fix for this in #183. The difference is that my solution looks at `black.__version__` and you're inspecting the actual signature of the function. Is this just a matter of taste or do you see pros and cons in each approach?

src/darker/black_diff.py
```diff
@@ -131,6 +131,9 @@ def apply_black_excludes(
     :return: Absolute paths of files which should be reformatted using Black
 
     """
+    sig = inspect.signature(gen_python_files)
+    # those two exist and are required in black>=21.8b0
```
akaihola, line 135: They were actually introduced in `21.7b1.dev9`.

src/darker/black_diff.py
```diff
@@ -131,6 +131,9 @@ def apply_black_excludes(
...
+    kwa = dict(verbose=False, quiet=False) if "verbose" in sig.parameters else {}
```
akaihola, line 136: Just a stylistic thing, I'd use the more commonly seen name `kwargs` for the extra keyword arguments.
```suggestion
    kwargs = dict(verbose=False, quiet=False) if "verbose" in sig.parameters else {}
```

src/darker/black_diff.py
```diff
@@ -141,6 +144,7 @@ def apply_black_excludes(
             force_exclude=black_config.get("force_exclude"),
             report=Report(),
             gitignore=None,
+            **kwa,
```
akaihola, line 147: ...here, too
```suggestion
            **kwargs,
```

akaihola:
Could you add a change log entry similar to [the one in the alternative pull request](https://github.com/akaihola/darker/pull/183/files#diff-ff3c479edefad986d2fe6fe7ead575a46b086e3bbcf0ccc86d85efc4a4c63c79R15)?

akaihola (changes requested):
You can see a couple of comments in the review. I'm starting to think I like the `inspect` approach better than looking at the Black version.

flying-sheep:
> Is this just a matter of taste or do you see pros and cons in each approach?

I think the approach here is simpler in two ways:

1. it uses less code
2. it checks directly for the difference that has to be handled as opposed to taking an indirect route over the version

but ultimately it’s a matter of taste as both solutions are likely to be robust.

akaihola (approved):
Thanks @flying-sheep for the improved solution!

PR #209, opened 2021-09-29
Performance improvements in __main__.py

### LINKED ISSUE #210: Loop invariant code slows down bisection when AST verification fails
As noted by @rogalski in #205,
>`run_black`, `diff_and_get_opcodes` and `opcodes_to_chunks` are run in each pass of bisect loop, despite being independent from context_lines variable: https://github.com/akaihola/darker/blob/d269159bd23391be7b0c414d550441a2438de718/src/darker/__main__.py#L150 https://github.com/akaihola/darker/blob/d269159bd23391be7b0c414d550441a2438de718/src/darker/__main__.py#L155 https://github.com/akaihola/darker/blob/d269159bd23391be7b0c414d550441a2438de718/src/darker/__main__.py#L158

These lines should be moved to be executed before the loop.

If AST verification fails, Darker starts to bisect in order to find the smallest possible number of extra diff context lines which preserves the AST. Moving loop invariant code outside the bisection loop will speed up this process.

rogalski:
Do not recompute invariant values in bisect loop.

Fixes #210.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -15,6 +15,7 @@ Fixed
 - ``/foo $ darker --diff /bar/my-repo`` now works: the current working directory can be
   in a different part of the directory hierarchy
 - An incompatible ``isort`` version now causes a short user-friendly error message
+- Improve bisect performance by not recomputing invariant data within bisect loop
 
 
 1.3.0_ - 2021-09-04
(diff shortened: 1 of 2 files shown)


akaihola:
@rogalski, there seem to be a couple of unfortunate trailing whitespaces. See [the build error](https://github.com/akaihola/darker/pull/209/checks?check_run_id=3759847425#step:5:57).

akaihola (changes requested):
Could you add a change log entry, and remove the trailing whitespaces caught by the CI build?

akaihola (approved):

PR #214, opened 2021-10-01
ASTVerifier

rogalski:
Improved performance by caching intermediate data during bisect.

Major changes:
- ASTVerifier
- fix up of tests to adjust implementation-aware mocks


src/darker/verification.py
```diff
@@ -67,3 +67,39 @@ def verify_ast_unchanged(
...
+            document_ast_str = self._to_ast_str(document)
+        except SyntaxError:
+            comparison = False
+        else:
+            comparison = self._baseline_ast_str == document_ast_str
+
+        self._comparisons[document.string] = comparison
+        return self._comparisons[document.string]
+
+    def assert_equivalent_to_baseline(self, document: TextDocument) -> None:
```
akaihola, line 102: `ASTVerifier.assert_equivalent_to_baseline()` is unused (except in tests). Should probably be removed?

src/darker/verification.py
```diff
@@ -67,3 +67,39 @@ def verify_ast_unchanged(
...
         debug_dump(black_chunks, edited_linenums)
         raise NotEquivalentError(str(exc_info))
+
+
+class ASTVerifier:
+    """
+    Verifies if reformatted TextDocument is AST-equivalent to baseline.
+
+    Keeps in-memory data about previous comparisons to improve performance.
+    """
```
akaihola, line 77: I'd like to use the same docstring convention everywhere: no newline before first line, end with a blank line.

```suggestion
    """Verify that reformatted TextDocument is AST-equivalent to baseline

    Keeps in-memory data about previous comparisons to improve performance.

    """
```

src/darker/tests/test_verification.py
```diff
@@ -28,6 +33,29 @@ def test_verify_ast_unchanged(dst_content, expect):
...
+    verifier = ASTVerifier(baseline=TextDocument.from_lines(["if True: pass"]))
+    assert verifier.is_equivalent_to_baseline(
+        TextDocument.from_lines(["if True:", "    pass"])
+    )
+    assert not verifier.is_equivalent_to_baseline(
+        TextDocument.from_lines(["if False: pass"])
+    )
+
+
+def test_ast_verifier_assert_equivalent():
```
akaihola, line 47: This test can go if the unused `assert_equivalent_to_baseline()` method is removed.

src/darker/verification.py
```diff
@@ -67,3 +67,39 @@ def verify_ast_unchanged(
...
+
+        try:
+            document_ast_str = self._to_ast_str(document)
+        except SyntaxError:
+            comparison = False
+        else:
+            comparison = self._baseline_ast_str == document_ast_str
+
+        self._comparisons[document.string] = comparison
+        return self._comparisons[document.string]
```
akaihola, line 100: There are two syntactically identical `return` statements. Would you consider the following a simplification and an improvement or not?
```suggestion
        if document.string not in self._comparisons:
            try:
                document_ast_str = self._to_ast_str(document)
            except SyntaxError:
                comparison = False
            else:
                comparison = self._baseline_ast_str == document_ast_str
            self._comparisons[document.string] = comparison
        return self._comparisons[document.string]
```

akaihola (changes requested):
Great implementation! Just a couple minor nitpicks...

src/darker/verification.py
rogalski: I left it because I was unsure if we guarantee stability of importable API.
If not, of course this can be removed.

src/darker/verification.py
rogalski: Sure, this is just trivial.

src/darker/tests/test_verification.py
rogalski: See my comment about backwards compatibility of importable API.

src/darker/verification.py
rogalski: This is purely opinion-based, but I prefer flatter structure and early returns.

Or maybe we should decouple caching and computing and just use `functools.lru_cache(max_size=None)`?

akaihola: I agree that double returns are a fair trade-off for flatness and an early return.

Separating caching would indeed be more clear, although the current version isn't bad either – caching can't get much simpler than this.

Is `@lru_cache` flushed when it's used on a method and the object goes out of scope? I mean if Darker is processing a huge repository of Python files with lots of changes, we wouldn't like it to keep this cache for files already processed. I didn't find an answer in [lru_cache documentation](https://docs.python.org/3/library/functools.html#functools.lru_cache).

akaihola: >Is @lru_cache flushed when it's used on a method and the object goes out of scope?

Hmm, apparently not:
```python
from functools import lru_cache

class CachedAdder:
    def __init__(self, a):
        self._a = a

    @lru_cache(maxsize=None)
    def add(self, b):
        return self._a + b
    
def test_adder():
    adder = CachedAdder(21)
    print(adder.add(21))
    
print(CachedAdder.add.cache_info())
test_adder()
print(CachedAdder.add.cache_info())
```
```
CacheInfo(hits=0, misses=0, maxsize=None, currsize=0)
42
CacheInfo(hits=0, misses=1, maxsize=None, currsize=1)
```

rogalski:
@akaihola 

Docstring fixes are pushed.
Mone missing UT for SyntaxError is comparison is pushed.

Comment about API deprecation vs. just removing code still stands.

About lru_cache - I decided to keep implementation as it is. Seems good enough to me.

We are close to getting that ready. Let's make sure we are on the same page regarding next steps.

src/darker/verification.py
akaihola: I don't get it, there hasn't been a `darker.verification.ASTVerifier` class before, so there's no existing stable API to keep here.

But the question about guaranteeing Python API stability is a valid one of course. I opened #216 for that discussion, thanks for bringing it up @rogalski!

PR #221, opened 2021-10-11
Experiment with getting reformatted chunks from Black

akaihola:
By getting results of reformatting from Black in chunks instead of complete files, we should be able to apply reformat modified sections with more granularity.

This is an experiment and to be considered carefully, since it involves re-implementing parts of Black, and will be sensitive to internal changes in Black versions.


akaihola:
I think I ran into a dead end and started to doubt whether this can be done at all.

PR #225, opened 2021-10-15
Return an exit value of `1` if any linter fails on modified lines

### LINKED ISSUE #223: Pre-Commit doesn't fail based on flake8 output
I am using flake8 as a linter by specifying it in pyproject.toml. Running `darker .` prints out the flake8 errors as expected.

However, when using darker in pre-commit hook, the Darker step passes. The output from flake8 is also not logged out. Is this expected behaviour? I was expecting it to fail.

For now, I am forced to use a separate hook for flake8 in order to fail the commit.

akaihola:
Fixes #223.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,8 @@ These features will be included in the next release:
 
 Added
 -----
+- Linter failures now result in an exit value of 1, regardless of whether ``--check``
+  was used or not. This makes linting in Darker compatible with ``pre-commit``.
 
 Fixed
 -----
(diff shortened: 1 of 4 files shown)


akaihola:
@jasleen19, would you by any chance have time to do a review of this pull request? I've invited you as one of the collaborators on this project, and if you accept, I'll be able to assign this for your review.

To test the fix, you should be able to use this branch with `pre-commit` by modifying your `.pre-commit-config.yaml` like so:
```yaml
- repo: https://github.com/akaihola/darker
  rev: linter-retval
  hooks:
  - id: darker
```

PR #238, opened 2021-11-04
isort only if imports at the top of the module were edited

### LINKED ISSUE #72: `--isort` sorts imports even if they were not edited
The problem with applying isort modifications "surgically" is in the fact that isort may (and often does) change the order of lines.

Could we consider all changes by isort as one contiguous diff chunk, and apply it if any line in `edited_linenums` falls inside that chunk? Is there an edge case where isort only modifies part of the import lines, and a user edit happens to an import outside that part? I can't see a problem there. The important thing is that either all or none of the isort modifications are applied.

akaihola:
Fixes #72:

>The problem with applying isort modifications "surgically" is in the fact that isort may (and often does) change the order of lines.
>
>Could we consider all changes by isort as one contiguous diff chunk, and apply it if any line in `edited_linenums` falls inside that chunk? Is there an edge case where isort only modifies part of the import lines, and a user edit happens to an import outside that part? I can't see a problem there. The important thing is that either all or none of the isort modifications are applied.

The changes are pretty involved. Here's a summary for reviewers:

### src/darker/\_\_main\_\_.py:
- some variable renames for clarity:
  - `path_in_repo` -> `relative_path_in_rev2`
  - `src` -> `absolute_path_in_rev2`
  - `relative_path` -> `relative_path_in_rev2`
- `_isort_and_blacken_single_file()` and `_isort_and_blacken_single_file()` now accept
  - a pre-created `EditedLinenumsDiffer` object instead of creating its own
  - a list of files to skip when reformatting using Black, instead of just a boolean for the single file

### src/darker/git.py:
- rename `get_rev1_path()` to `get_path_in_repo()`

### src/darker/import_sorting.py:
- `apply_isort()` now
  - accepts a pre-created `EditedLinenumsDiffer` object instead of creating its own
  - calls into helper functions refactored out of it:
    - `isort` arguments are created in `_build_isort_args()`
    - `isort` is invoked in `_call_isort_code()`
  - generates a diff for the result of `isort`
- `_diff_overlaps_with_edits()` does the meat of the new functionality: it checks whether the range of user edits touches the range of reformatted imports

### src/darker/diff.py:
- `diff_chunks()` now combines `diff_and_get_opcodes()` and `opcodes_to_chunks()` behind one call

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,7 +8,9 @@ Added
 - The ``--workers``/``-W`` option now specifies how many Darker jobs are used to
   process files in parallel to complete reformatting/linting faster.
 - Linters can now be installed and run in the GitHub Action using the ``lint:`` option.
-  
+- Sort imports only if the range of modified lines overlaps with changes resulting from
+  sorting the imports.
+
 Fixed
 -----
 - Avoid memory leak from using ``@lru_cache`` on a method.
--- a/src/darker/__main__.py
+++ b/src/darker/__main__.py
@@ -125,7 +125,7 @@ def _isort_and_blacken_single_file(  # pylint: disable=too-many-arguments
 
     """
     # With VSCode, `relative_path_in_rev2` may be a `.py.<HASH>.tmp` file in the
-    # working tree insted of a `.py` file.
+    # working tree instead of a `.py` file.
     absolute_path_in_rev2 = root / relative_path_in_rev2
     rev2_content = git_get_content_at_revision(
         relative_path_in_rev2, revrange.rev2, root
--- a/src/darker/__main__.py
+++ b/src/darker/__main__.py
@@ -136,7 +136,7 @@ def _isort_and_blacken_single_file(  # pylint: disable=too-many-arguments
         content_after_reformatting = _blacken_single_file(
             root,
             relative_path,
-            revrange,
+            EditedLinenumsDiffer(root, revrange),
             rev2_content,
             rev2_isorted,
             enable_isort,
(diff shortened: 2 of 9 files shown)


akaihola:
@Yajo, I noticed you gave a thumbs up for this idea in #72. I could include this change in [release 1.5.0](https://github.com/akaihola/darker/milestone/13) – would you be able to do a code review of my pull request here?

akaihola:
I hopefully managed to clarify the changes a bit in the description of this PR to make reviewing easier. @roniemartinez would you happen to have a bit of time to do a sanity check on this patch?

Yajo:
I've got no idea about the code sorry 😕 




However I've tested this and it doesn't seem to do any isort at all. I'm not sure if I'm testing well, though. AFAICS only black runs:

https://user-images.githubusercontent.com/973709/162377838-11c4c305-9589-4d7b-913f-84523a06e4d2.mp4

akaihola:
@Yajo, thanks for testing this! You do need the `-i`/`--isort` option to sort imports as well. It's not enabled by default.

akaihola:
@roniemartinez, if the review feels too messy, I could also separate out e.g. the variable renames, function refactorings and `EditedLinenumsDiffer` changes into their own PRs which don't change functionality. Would this help?

roniemartinez (approved):
LGTM

akaihola:
Thanks @roniemartinez! We're getting close to a [1.5.0 release](https://github.com/akaihola/darker/milestone/13)!

PR #255, opened 2022-01-18
VSCode `.py.tmp` file should be reformatted

### LINKED ISSUE #249: tmp files filtered out before checking for matching tracked file
Noticed the problem when configuring Darker on VSCode. Running the formatting command explicitly works but formatting on save does not. As noted in other issues, the only difference in executed command is that formatting on save runs on a tmp file (*.hash.tmp) instead of the file directly. Support was added to check for these tmp files but the check for them happens after filtering out the file.

The root of the error I believe is in calling `git_get_modified_files`, which in turn calls `should_reformat_file` which only allows `.py` files. This happens before ever accounting for possible tmp files.

I will poke a bit and see if just adding the check to `should_reformat_file` is sufficient to get things working.

## Software Versions
- Darker: 1.3.2
- Python: 3.9.9
- Black: 21.12b0
- Isort: 5.10.1
- VSCode: 1.63.2

akaihola:
Fixes #249 and #255

- [x] rebase on `master` after #270 has been merged
- [x] failing test
- [x] fix
- [x] change log

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,8 @@ Added
 
 Fixed
 -----
+- Consider ``.py.tmp`` as files which should be reformatted.
+  This enables VSCode Format On Save.
 
 
 1.4.0_ - 2022-02-08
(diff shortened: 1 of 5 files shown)


akaihola:
@magnunm this is now ready for review.

magnunm (approved):
Looks good

PR #257, opened 2022-01-19
No additional newline with `--stdout` flag.

### LINKED ISSUE #250: Output with `--stdout` flag contains an additional newline compared to without the flag
Compared to the result of modifying the file, the output with `--stdout` contains an additional newline at the end of the file. As I set up my editor to use the stdout to format the file, this would add a extra newline when I run the formatting.

The cause is the `print` statements here: [\_\_main\_\_.py#L257-L270](https://github.com/akaihola/darker/blob/c5d0f68f410d5cd9b3998627da2171695cf4941b/src/darker/__main__.py#L257-L270), as `print` adds a newline at the end. Have changed it to the way I prefer in my local copy by adding `end=""` to those print statements. Unless the extra newline is in fact wanted, I can make a PR with that change.

**How to reproduce:**
In an initialized, empty project, run:
```shell
$ echo "print('Some test code')" > test.py
$ cat test.py
print('Some test code')
$ darker --stdout test.py
print("Some test code")

$ darker test.py && cat test.py
print("Some test code")
```

**Environment**:
darker: commit c5d0f68f410d5cd9b3998627da2171695cf4941b
python: 3.10
shell: GNU bash

magnunm:
Fixes #250

- [x] rebase on `master` after #270 has been merged

### DIFF
--- a/src/darker/tests/test_main.py
+++ b/src/darker/tests/test_main.py
@@ -8,7 +8,7 @@ from argparse import ArgumentError
 from pathlib import Path
 from textwrap import dedent
 from types import SimpleNamespace
-from unittest.mock import call, patch
+from unittest.mock import Mock, call, patch
 
 import pytest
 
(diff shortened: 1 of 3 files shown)


akaihola:
The `test_stdout_path_resolution()` unit test now fails as expected. And actually in that unit test we can see that the behavior was indeed wrong!

I'm also noticing that we didn't really have tests for the output behavior of `--stdout`. I think we need them, both for output to a TTY with and without Pygments, and for non-TTY output. Any idea how to set up those three cases?

akaihola:
Ah, and the `colorize` test failure is probably due to a change in a new version of Pygments, and I'm fixing that in another branch.

magnunm:
Will see what I can do about the `--stdout` tests!

magnunm:
@akaihola Made an attempt in 56626f393602f044e096eca7aba04b144afa18a5, will look at it again later. Do you understand the Python <= 3.8 failure? 
The `pygments` case also seems flaky, the two different variants are what I get running the whole test-suite and just the `test_main.py`... Maybe it is better to `expect=highlight(...`?

akaihola:
Could you rebase on `master` since I now merged #258?

I'll take a look at the py<3.8 test failures next. Maybe a context manager syntax compatibility issue?

src/darker/tests/test_main.py
```diff
@@ -775,11 +775,58 @@ def test_modify_file(tmp_path, new_content, expect):
...
+def test_print_source(new_content, tty, with_pygments, expect, capsys):
+    """Highlight is applied only if tty, final newline is handled correctly."""
+    with (
+        patch("sys.stdout.isatty", Mock(return_value=tty)),
+        patch(
+            "darker.__main__._import_pygments",
+            Mock(
+                return_value=darker.__main__._import_pygments(),
+                side_effect=None if with_pygments else ImportError(),
+            ),
```
akaihola, line 818: Parenthesized context managers are a recent addition to Python syntax. We should support older versions using this old uglier syntax (made even worse by Black):
```suggestion
    with patch("sys.stdout.isatty", Mock(return_value=tty)), patch(
        "darker.__main__._import_pygments",
        Mock(
            return_value=darker.__main__._import_pygments(),
            side_effect=None if with_pygments else ImportError(),
```

akaihola (approved):
This looks great, ready to merge in my opinion!

PR #267, opened 2022-01-31
fix support for black 22.1

jabesq:


src/darker/config.py
```diff
@@ -104,7 +104,8 @@ def load_config(srcs: Iterable[str]) -> DarkerConfig:
                  are used to look for the ``pyproject.toml`` configuration file.
 
     """
-    path = find_project_root(tuple(srcs or ["."])) / "pyproject.toml"
+    project_root = find_project_root(tuple(srcs or ["."]))[0]
+    path = project_root / "pyproject.toml"
```
Mystic-Mirage, line 108: Let's use `find_pyproject_toml` instead:

```python
    path = find_pyproject_toml(tuple(srcs or ["."]))
    if path is not None:
```

akaihola: What if someone uses Darker with a Python code base which doesn't have `pyproject.toml`?

akaihola:
As noted elsewhere, in #270 I took the approach of supporting both the old and the new Black behavior.

akaihola:
#270 has been merged instead. Thank you @jabesq for your contribution nevertheless!

PR #269, opened 2022-02-01
Add support for `black` >= 22.1.0

Mystic-Mirage:
Black changed `find_project_root` behavior. Now it returns a tuple.
Let's use `find_pyproject_toml` instead.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -26,7 +26,7 @@ package_dir =
     =src
 packages = find:
 install_requires =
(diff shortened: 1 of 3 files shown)


Mystic-Mirage:
Duplicates #267

akaihola:
I took yet another approach to this in #270 by accepting both the old and the new behavior, and encapsulating that in a "Black compatibility" module.

PR #279, opened 2022-02-07
Drop support for Python 3.6

akaihola:


MatthijsBurgh (approved):
Conflicting with #288

MatthijsBurgh (approved):

MatthijsBurgh:
I would say that the version constraints of `request_cache` should be `>=0.7`. The upper limit should be removed, but I don't see a reason the increase the lower limit.

MatthijsBurgh (approved):

akaihola:
@MatthijsBurgh, I added one more change to run `pyupgrade` in the GitHub workflow with Python 3.7 as the minimum, and fixed the syntax to be py37+ in two places.

PR #282, opened 2022-02-10
Update the GitHub Action to use Darker 1.4.0

### LINKED ISSUE #260: Darker GH action checks against HEAD commit -> no diff
The Darker GitHub action is not working. As without specifying the revision argument, it checks against HEAD.

So in GH actions, it will always check against the current HEAD, which means the diff is always empty. Of course I could set up some custom logic to get the revision for a `push` or `pull request` in my workflow. But it would be really helpful for everybody if would be included in the action.

So the cases to consider in my opinion are just the push and PR events for now:
Push event: Get the diff compared to last run (on this branch)
PR event: Get the diff for the entire PR.
For my own CI, I have used the following logic, maybe you could use it: https://github.com/tue-robotics/tue-env/blob/master/ci/commit-range/action.yml (In case `COMMIT_RANGE` is empty, I go back to `HEAD^`. This should only cause issues when HEAD is the first commit in the repo.)

akaihola:
Fixes #260.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,7 +11,8 @@ Fixed
 -----
 - Consider ``.py.tmp`` as files which should be reformatted.
   This enables VSCode Format On Save.
-
+- Use Darker 1.4.0 instead of 1.3.2 in the GitHub Action.
+  
 
 1.4.0_ - 2022-02-08
 ===================
--- a/README.rst
+++ b/README.rst
@@ -544,7 +544,7 @@ Create a file named ``.github/workflows/darker.yml`` inside your repository with
              fetch-depth: 0 
          - uses: akaihola/darker@1.4.0
            with:
-             options: "--check --diff"
+             options: "--check --diff --revision=master..."
              src: "./src"
              version: "1.4.0"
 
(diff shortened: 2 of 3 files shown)


MatthijsBurgh (approved):

MatthijsBurgh:
CI probably fails because of an update in `types-toml`
Total diff in installed dependencies to last working run.
```
--- <unnamed>
+++ <unnamed>
@@ -13,7 +13,7 @@
 mypy-extensions-0.4.3
 packaging-21.3
 pathspec-0.9.0
-platformdirs-2.4.1
+platformdirs-2.5.0
 pluggy-1.0.0
 py-1.11.0
 pycodestyle-2.7.0
@@ -30,6 +30,6 @@
 regex-2022.1.18
 toml-0.10.2
 tomli-2.0.1
-types-toml-0.10.3
+types-toml-0.10.4
 typing-extensions-4.0.1
 wrapt-1.13.3
```

akaihola:
Thanks @MatthijsBurgh, that's probably the case. I have a fix in #283 and will rebase this one on `master` once that one is merged.

PR #284, opened 2022-02-12
Run a CI test against Black `main` branch

### LINKED ISSUE #186: Changes in Black internals may break Darker
We're using Black's Python API directly. Black doesn't yet define a public API (see psf/black#779), so changes in Black may break Darker. This has in fact already happened at least twice.

To get an early warning about such changes, we could add a GitHub Action workflow for testing against Black `main` instead of its latest release.

akaihola:
Fixes #186.

@yoursvivek would you be able to review this? Would you accept an invitation as a collaborator on this repository?


yoursvivek (commented):
Reviewing post merge, but change looks good to me.

> Would you accept an invitation as a collaborator on this repository?

I'll be glad to help out. Count me in.

akaihola:
Yes @yoursvivek, I figured I just had to merge it in order to see whether a cron schedule workflow runs on `master`. I tried to schedule it for Saturdays at 20:05 UTC (ten minutes ago), but it didn't run. I wonder why.

But anyway, the "future" test also runs as part of pull requests, and that seems to work fine.

PR #296, opened 2022-02-18
Include `isort@main` in `test-future` workflow

### LINKED ISSUE #262: Pin `isort` in the GitHub Action image

akaihola:
Fixes #262.

Also adds use of `--upgrade --upgrade-strategy=eager` for pip when installing packages with `constraints-future.txt`. This way not only Black and `isort` are from their `main` branches, but also we get newest releases for all other dependencies as well.


akaihola:
`isort@main` was cloned successfully in [build #847](https://github.com/akaihola/darker/runs/5244290965?check_suite_focus=true).

Issue #47, opened 2020-08-08
Switch from argparse to click


akaihola:
Black uses it anyway and is a dependency of Darker.

Does [click](https://pypi.org/project/click/) allow limiting arguments to only one occurrence?

[Argparse](https://docs.python.org/3/library/argparse.html) doesn't support that out of the box, so
```bash
$ myscript.py -a first -a second -a third
```
will either ignore all but `-a third`, or collect them all to a list (with `action="append"`).

Carreau:
> Does [click](https://pypi.org/project/click/) allow limiting arguments to only one occurrence?

No, it does like argparse and `--a A --a B` will end up with `B` as the value. You can use multiple=True and check the length.

```
import click

@click.command()
@click.option('--name', nargs=1, multiple=False)
def hello(name):
    click.echo(f"Hello, {name}!")

if __name__ == '__main__':
    hello()
```

Carreau:
I'm also not sure click will be that simpler, I find that click is useful if you have many subcommands. Not that much when you only have one with many options.

akaihola:
I've also been quite comfortable with argparse most of the time, and happy to not bring yet another dependency into projects.

So let's postpone this change until we really need something that argparse can't do easily.

PR #298, opened 2022-02-18
Document VSCode / `--lint` incompatibility

### LINKED ISSUE #297: Emit warning or error if linting is enabled and `*.py.<hash>.tmp` is being reformatted
It seems like the problem in #286 was that Darker was configured to run linters while VSCode only expects to see a reformatting diff. Darker could
- [ ] complain about this in a suitable way which would catch the attention of the VSCode user
- [ ] document this limitation.

See also #252 and #254.

akaihola:
Fixes #297 partially.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -639,7 +639,7 @@ Here's an example of `cov_to_lint.py`_ output::
 +=======================================================================+
 | Don't enable linting on the command line or in the configuration when | 
 | running Darker as a reformatter in VSCode. You will confuse VSCode    |
-| with unexpected output from Darker.                                   |
+| with unexpected output from Darker, as it only expect black's output                                  |
 +-----------------------------------------------------------------------+
 
 .. _Mypy: https://pypi.org/project/mypy
--- a/README.rst
+++ b/README.rst
@@ -311,6 +311,10 @@ For example:
    ]
    log_level = "INFO"
 
+Be careful to not use options which generate output which is unexpected for
+other tools. For example, VSCode only expects the reformat diff, so
+``lint = [ ... ]`` can't be used with it.
+
 *New in version 1.0.0:*
 
 - The ``-c``, ``-S`` and ``-l`` command line options.
@@ -413,6 +417,7 @@ PyCharm/IntelliJ IDEA
 
 __ https://plugins.jetbrains.com/plugin/7177-file-watchers
 
+
 Visual Studio Code
 ------------------
 
@@ -440,6 +445,8 @@ Visual Studio Code
 
 You can pass additional arguments to ``darker`` in the ``blackArgs`` option
 (e.g. ``["--diff", "--isort"]``), but make sure at least ``--diff`` is included.
+Also, this won't work if you have any linters enabled on the command line
+or in ``pyproject.toml``.
 
 Note that VSCode first copies the file to reformat into a temporary
 ``<filename>.py.<hash>.tmp`` file, then calls Black (or Darker in this case) on that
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -438,6 +443,8 @@ Visual Studio Code
 
 You can pass additional arguments to ``darker`` in the ``blackArgs`` option
 (e.g. ``["--diff", "--isort"]``), but make sure at least ``--diff`` is included.
+Also, this won't work if you have any linters enabled on the command line
+or in ``pyproject.toml``.
```
mayk0gan, line 447: that's not true though, it seems that if you don't specify `diff`/`check` in the `pyproject.toml` file - you can override its settings in the cli. (If you do specify them, it doesn't allow to override its configuration).

Though I haven't checked why

README.rst
```diff
@@ -625,6 +632,14 @@ Here's an example of `cov_to_lint.py`_ output::
     src/darker/__main__.py:131: no coverage:                 raise
     src/darker/__main__.py:132: no coverage:             logger.debug(
 
++-----------------------------------------------------------------------+
+|                               ⚠ NOTE ⚠                                |
++=======================================================================+
+| Don't enable linting on the command line or in the configuration when | 
+| running Darker as a reformatter in VSCode. You will confuse VSCode    |
+| with unexpected output from Darker.                                   |
```
mayk0gan, line 640:
```suggestion
| with unexpected output from Darker, as it only expect black's output                                  |
```

mayk0gan (changes requested):

README.rst
akaihola: Yes, including a binary option (like `--diff`, `--check` or `--isort`) in
- just `pyproject.toml`
- just on the command line
- both in `pyproject.toml` and the command line

will enable that option. Is that what you meant?

Also, once enabled in either place, it's currently not possible to override and disable it in the other.

What I mean with "won't work if linters enabled" is that VSCode will not be able to apply the reformattings if there are any linter errors/warnings mixed in Darker's output. Is there a better way to express that?

Also, it seems to me that `--diff --quiet` are hard-coded as options for Black in VSCode, and anything added to `blackArgs` will be appended to those. So that we could fix here as well.

akaihola: I clarified this whole paragraph. Take a look at the newest version, is it now correct and more understandable?

README.rst
```diff
@@ -627,6 +635,14 @@ Here's an example of `cov_to_lint.py`_ output::
     src/darker/__main__.py:131: no coverage:                 raise
     src/darker/__main__.py:132: no coverage:             logger.debug(
 
++-----------------------------------------------------------------------+
+|                               ⚠ NOTE ⚠                                |
++=======================================================================+
+| Don't enable linting on the command line or in the configuration when | 
+| running Darker as a reformatter in VSCode. You will confuse VSCode    |
+| with unexpected output from Darker, as it only expect black's output                                  |
```
mayk0gan, line 643:
```suggestion
| with unexpected output from Darker, as it only expect black's output  |
```

mayk0gan (changes requested):
just one little thing and you can merge. Thanks for the clarification :)

README.rst
mayk0gan: Great clarification, thanks.
Is there any reason that boolean-arguments are non overridable? Seems like we should open an issue on that

akaihola: @mayk0gan I agree, boolean options should be overridable. We could also switch from [argparse](https://docs.python.org/3/library/argparse.html) to [click](https://click.palletsprojects.com/) (see #47) to make that easier.

PR #303, opened 2022-02-21
Skip linting in certain CI configurations

akaihola:
I extracted these changes from #179 (`--jobs argument`) since it seems merging that one is still getting delayed.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -58,12 +58,10 @@ jobs:
           - os: ubuntu-latest
             python-version: '3.7'
             constraints: '--constraint constraints-oldest.txt'
-            lint: ''
           - os: ubuntu-latest
             python-version: '3.10'
             constraints: '--constraint constraints-future.txt'
             upgrade: '--upgrade --upgrade-strategy=eager'
-            lint: ''
     env:
       RUN_MYPY: >-
         ${{ matrix.python-version == '3.9' || matrix.python-version == '3.10' }}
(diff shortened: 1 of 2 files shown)


.github/workflows/python-package.yml
```diff
@@ -75,6 +80,11 @@ jobs:
           # strict dependency resolution added in pip 20.3
           python -m pip install --upgrade 'pip>=20.3'
           pip install -e '.[isort,test]' ${{ matrix.constraints }}
-      - name: Test with pytest
+      - name: Run Pytest without Mypy if Python < 3.9
+        if: env.RUN_MYPY == 'false'
```
MatthijsBurgh, line 84: I think you could do the python version check here

akaihola: I'd like to avoid repeating the `matrix.python-version == '3.9' || matrix.python-version == '3.10'` check twice. Can you think of a way to do that if the check is done here instead of in `env:` above?

MatthijsBurgh: We can reduce the version check to be `>= 3.9`. I think that is also more future proof. And only change it in the future when it is necessary.
But will need to have two check. Unless you putt all the logic in the `run` of one step.

akaihola: I believe `matrix.python-version >= '3.9'` will be false for `matrix.python-version == '3.10'`
(since the version numbers need to be strings). So we'd need to repeat the
`matrix.python-version == '3.9' || matrix.python-version == '3.10'` comparison twice. Which is why I separated it.

Instead of a `'true'`/`'false'`, we could produce `'--mypy'`/`''` and merge the two `run`s
into one `pytest ${{ matrix.lint }} ${{ env.RUN_MYPY }}`. Do you think that would be an improvement?

akaihola: I can't think of a better way currently. I'll keep this as is now and merge this PR. Thanks for your comments @MatthijsBurgh!

Issue #304, opened 2022-02-22
[DESIGN] support different code formatters, allow to skip formatting


akaihola:
See also: [many-formatters](/users/akaihola/projects/4/views/5) sub-project

Currently there's no way to run Darker and have it just do, say, `isort` and `flynt` but *not* reformat code using Black.

Also, there are open requests for replacing Black with another formatter or expanding the choices of formatters/linters:
- #51 
- #251 / #309
- #266 
- #521 

We could expand Darker's behavior in stages:
- extend configuration parameters:
  - add `--black`/`--no-black`; see
    - #449
    - #301
  - OR: add `--format=black`/`--format=blue`/`--format=ruff`/`--format=none`
    - see #562
- add support for `blue`, `Ruff` and other formatters; see
  - #309
  - #521
- add another console entry point which defaults to `--format=none` (what would be a good name for that?)
- change default for `darker` to `--format=none` in version 2.0.0
  - see #450

For the option name, I've considered `--format`, `--formatter`, `--reformat` and `--reformatter`:
- Black calls itself a "formatter", so I discarded `--reformat` and `--reformatter` (also for needless verbosity)
- Linters are chosen with `--lint`, not `--linter`, so for consistency, I'd prefer `--format` over `--formatter`

At some point after making `--format=none` the default, `black` should be removed as a hard dependency. I can see two alternatives:
- `pip install 'darker[black]'` and `darker --black` for the current behavior
- split into two Python packages:
  - move all code from `darker` to a new package (package name to be decided) which
    - doesn't require `black`
    - doesn't run `black` by default (so has `--format=none` as the default)
  - make `darker` depend on the new package and `black`
  - in `darker`, wrap and command line parser of the new package and override `--format` by setting `--format=black` as default

For the `--black`/`--no-black`/`--blue`/`--no-blue` design choice, some pros and cons are:
- :green_circle::+1:  `--black`: can enable multiple formatters (although what about the running order?)
- :green_circle::+1:  `--black`: consistent with `--isort` and `--flynt`
- :green_circle::+1:  `--black`: shorter to type

and for `--format={black|blue|none}`:
- :green_circle::+1:  `--format=`: consistent with `--lint=`
- :red_circle::-1: `--format=` inconsistent with `--lint=` since `--format=black` or `--format=blue` don't really specify a command line
- :red_circle::-1: `--format=`: more verbose

akaihola:
[...]

PR #309, opened 2022-02-24
Add Blue as an alternative reformatter

### LINKED ISSUE #251: `blue` instead of `black`?
Is it possible to use [blue](https://blue.readthedocs.io/en/latest/) instead of black?

`blue` is a somewhat less uncompromising code formatter than `black`.

akaihola:
Fixes #251.

- [x] Merge #738 
- [ ] Implement Blue as a plugin
- [ ] Add tests for the Blue plugin

**Additional context**
- [many-formatters](/users/akaihola/projects/4/views/5) sub-project
- #304
- #450 
- #449

### DIFF
--- a/constraints-oldest.txt
+++ b/constraints-oldest.txt
@@ -5,6 +5,7 @@
 # versions in `setup.cfg`.
 black==22.3.0
 darkgraylib==2.2.0
+blue==0.8.0
 defusedxml==0.7.1
 flake8-2020==1.6.1
 flake8-bugbear==22.1.11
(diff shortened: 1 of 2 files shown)


akaihola:
Blue seems to be dead. The latest [0.9.1] version was released in August 2022, and the latest commit to [grantjenks/blue] was in October 2022. The internal Python API is not identical to Black versions supported by Darker, so some work would be needed. This is probably not worth the effort.

[0.9.1]: https://pypi.org/project/blue/0.9.1/
[grantjenks/blue]: /grantjenks/blue

PR #310, opened 2022-02-24
Check Darker code with `bandit`, `codespell`, `pyupgrade` and `safety`

cclauss:
Test results: https://github.com/cclauss/darker/actions


akaihola:
Thanks @cclauss for your pull request!

I'd like to discuss the thinking behind it in more detail, though. Could you explain the motivation behind adding all these extra checks on Darker's code base?

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,25 @@
...
+  lint_python:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - run: pip install --upgrade pip wheel
+      - run: pip install bandit codespell darker flake8 flake8-2020 flake8-bugbear
+                         flake8-comprehensions isort mypy pytest pyupgrade safety
+      - run: bandit --recursive --skip B101 . || true  # B101 is assert statements
+      - run: darker --check .
```
akaihola, line 13: We already run Darker itself on its own code base using [pytest-darker](https://pypi.org/project/pytest-darker/), see
- [`.github/workflows/python-pacakge.yml`](https://github.com/akaihola/darker/blob/8a1835c952b9743ea5cb6d40d23eb9c6458e7828/.github/workflows/python-package.yml#L81-L83)
- [`setup.cfg`](https://github.com/akaihola/darker/blob/8a1835c952b9743ea5cb6d40d23eb9c6458e7828/setup.cfg#L58)
- [`pytest.ini`](https://github.com/akaihola/darker/blob/8a1835c952b9743ea5cb6d40d23eb9c6458e7828/pytest.ini#L5)

I agree it might be a better idea to run Darker explicitly instead of through [pytest-darker](https://pypi.org/project/pytest-darker/). In that case, the PR that makes the change should also remove `--darker` from [`pytest.ini`](https://github.com/akaihola/darker/blob/8a1835c952b9743ea5cb6d40d23eb9c6458e7828/pytest.ini#L5).

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,25 @@
...
+      - run: codespell || true  # --ignore-words-list="" --skip="*.css,*.js,*.lock"
+      - run: flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+      - run: flake8 . --count --exit-zero --max-complexity=10 --max-line-length=88
+                      --show-source --statistics
```
akaihola, line 18: We also already run Flake8 on using [pytest-darker](https://pypi.org/project/pytest-darker/), see
- [`pytest.ini`](https://github.com/akaihola/darker/blob/8a1835c952b9743ea5cb6d40d23eb9c6458e7828/pytest.ini#L6)
- [`setup.cfg`](https://github.com/akaihola/darker/blob/8a1835c952b9743ea5cb6d40d23eb9c6458e7828/setup.cfg#L68-L76) (`[flake8]` configuration section)

If we were to run Flake8 explicitly instead of through [pytest-darker](https://pypi.org/project/pytest-darker/), we should also remove `--flake8` from [`pytest.ini`](https://github.com/akaihola/darker/blob/8a1835c952b9743ea5cb6d40d23eb9c6458e7828/pytest.ini#L6). Maybe also configure Flake8 in [`setup.cfg`](https://github.com/akaihola/darker/blob/8a1835c952b9743ea5cb6d40d23eb9c6458e7828/setup.cfg#L68-L76) (`[flake8]` configuration section) rather than using command line arguments?

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,25 @@
...
+      - run: isort --check-only --profile black . || true
```
akaihola, line 19: `isort` is already being run using [pytest-darker](https://pypi.org/project/pytest-darker/) as well, see
- [`pytest.ini`](https://github.com/akaihola/darker/blob/8a1835c952b9743ea5cb6d40d23eb9c6458e7828/pytest.ini#L7)
- [`pyproject.toml`](https://github.com/akaihola/darker/blob/8a1835c952b9743ea5cb6d40d23eb9c6458e7828/pyproject.toml#L8-L10) (`[tool.isort]` configuration section)

If we'd run `isort` explicitly instead of through [pytest-darker](https://pypi.org/project/pytest-darker/), I'd remove `--isort` from [`pytest.ini`](https://github.com/akaihola/darker/blob/8a1835c952b9743ea5cb6d40d23eb9c6458e7828/pytest.ini#L7), and rely on [`pyproject.toml`](https://github.com/akaihola/darker/blob/8a1835c952b9743ea5cb6d40d23eb9c6458e7828/pyproject.toml#L8-L10) for `isort` configuration instead of command line options here. This way running `isort` or `darker --isort` manually would use the same options.

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,25 @@
...
+      - run: pip install -r requirements.txt || pip install --editable . || true
+      - run: mkdir --parents --verbose .mypy_cache
+      - run: mypy --ignore-missing-imports --install-types --non-interactive . || true
```
akaihola, line 22: `mypy` is already being run using [pytest-darker](https://pypi.org/project/pytest-darker/) as well, see
- [`pytest.ini`](https://github.com/akaihola/darker/blob/8a1835c952b9743ea5cb6d40d23eb9c6458e7828/pytest.ini#L8)
- [`mypy.ini`](https://github.com/akaihola/darker/blob/8a1835c952b9743ea5cb6d40d23eb9c6458e7828/mypy.ini)

And again, if we'd run `mypy` explicitly instead of through [pytest-darker](https://pypi.org/project/pytest-darker/), `--mypy` needs to be removed from [`pytest.ini`](https://github.com/akaihola/darker/blob/8a1835c952b9743ea5cb6d40d23eb9c6458e7828/pytest.ini#L8). At least the `--ignore-missing-imports` option should be enabled per package in [`mypy.ini`](https://github.com/akaihola/darker/blob/8a1835c952b9743ea5cb6d40d23eb9c6458e7828/mypy.ini) (as it's already being done) instead of using the command line options. This way running `mypy` or `darker --mypy` manually would give the same results as the CI run.

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,25 @@
...
+      - run: pytest . || true
+      - run: pytest --doctest-modules . || true
```
akaihola, line 24: `pytest` is already being run in [`.github/workflows/python-pacakge.yml`](https://github.com/akaihola/darker/blob/8a1835c952b9743ea5cb6d40d23eb9c6458e7828/.github/workflows/python-package.yml#L81-L83). It is configured in [`pytest.ini`](https://github.com/akaihola/darker/blob/8a1835c952b9743ea5cb6d40d23eb9c6458e7828/pytest.ini) to run `--doctest-modules` as well.

Note that if you run both
```bash
pytest .
```
and
```bash
pytest --doctest-modules .
```
you're actually discovering and running unit tests twice!

```suggestion
```

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,25 @@
...
+      - run: pip install bandit codespell darker flake8 flake8-2020 flake8-bugbear
+                         flake8-comprehensions isort mypy pytest pyupgrade safety
+      - run: bandit --recursive --skip B101 . || true  # B101 is assert statements
+      - run: darker --check .
+      - run: codespell || true  # --ignore-words-list="" --skip="*.css,*.js,*.lock"
+      - run: flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+      - run: flake8 . --count --exit-zero --max-complexity=10 --max-line-length=88
+                      --show-source --statistics
+      - run: isort --check-only --profile black . || true
+      - run: pip install -r requirements.txt || pip install --editable . || true
```
akaihola, line 16: Darker has no `requirements.txt` file. Also, surely you'd like the workflow to fail if installation of the package fails in the first place.

```suggestion
      - run: pip install --editable .
```

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,25 @@
...
+      - run: mkdir --parents --verbose .mypy_cache
```
akaihola, line 17: I believe `mypy` creates the cache directory automatically if it's missing. At least I've never had to create it manually.
```suggestion
```

cclauss: In a GitHub Action or on your local disk?

.github/workflows/lint_python.yml
cclauss: error: --non-interactive is only supported with --install-types

cclauss:
What versions of Python are supported?  Is Python 3.6 supported er not?

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,25 @@
...
+jobs:
+  lint_python:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - run: pip install --upgrade pip wheel
+      - run: pip install bandit codespell darker flake8 flake8-2020 flake8-bugbear
+                         flake8-comprehensions isort mypy pytest pyupgrade safety
+      - run: bandit --recursive --skip B101 . || true  # B101 is assert statements
```
akaihola: `bandit` is indeed a good addition. Maybe it would make sense to run it twice: once for actual code and once for the tests directory?

Or, if it's possible to configure skipped checks individually for each module (or even class/function), I'd opt for that. Do you @cclauss know whether that's supported in `bandit`?

Also:
- `B101` shouldn't be skipped outside tests, since there shouldn't be any asserts in the actual package
- Darker relies on `subprocess` to run Git. `B404` and `B603` should be skipped in some modules.
- We do have this in `darker/tests/test_linting.py`:
  ```python
  # Use evil `eval()` so we get Windows compatible expected paths:
  # pylint: disable=eval-used
  assert result == [
      eval(f'f"{line}"', {"git_repo": git_repo}) for line in expect_output
  ]
  ```
  and it triggers `B307` which recommends `ast.literal_eval()`. This is a good heads-up and should probably be looked into. Would you @cclauss happen to have time to think about a better alternative for this test case while keeping Windows compatibility?

.github/workflows/lint_python.yml
cclauss: It is to see if there are doctests er not to pick the right one to run longterm.

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,25 @@
...
+      - run: darker --check .
+      - run: codespell || true  # --ignore-words-list="" --skip="*.css,*.js,*.lock"
```
akaihola: Wow, `codespell` is cool and actually uncovered two actual spelling mistakes!

This would have to skip `.git/` and `.mypy_cache` though.

```suggestion
      - run: codespell --skip="*.json,.git" || true
```

cclauss: The incredible thing about codespell is its speed even on a massive codebase.

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,25 @@
...
+      - run: flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+      - run: flake8 . --count --exit-zero --max-complexity=10 --max-line-length=88
+                      --show-source --statistics
+      - run: isort --check-only --profile black . || true
+      - run: pip install -r requirements.txt || pip install --editable . || true
+      - run: mkdir --parents --verbose .mypy_cache
+      - run: mypy --ignore-missing-imports --install-types --non-interactive . || true
+      - run: pytest . || true
+      - run: pytest --doctest-modules . || true
+      - run: shopt -s globstar && pyupgrade --py36-plus **/*.py || true
```
akaihola: I would prefer the workflow to fail in case `pyupgrade` fails.

```suggestion
      - run: shopt -s globstar && pyupgrade --py36-plus **/*.py
```

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,25 @@
...
+on: [pull_request, push]
+jobs:
+  lint_python:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - run: pip install --upgrade pip wheel
+      - run: pip install bandit codespell darker flake8 flake8-2020 flake8-bugbear
+                         flake8-comprehensions isort mypy pytest pyupgrade safety
```
akaihola: All these packages should be added to [`setup.cfg`](https://github.com/akaihola/darker/blob/8a1835c952b9743ea5cb6d40d23eb9c6458e7828/setup.cfg#L51-L66) in the `test` extras dependencies, and Darker development + test dependencies should be installed already here using:
```suggestion
      - run: pip install --editable '.[test]'
```

akaihola:
> What versions of Python are supported? Is Python 3.6 supported er not?

Python 3.6 support will be dropped in Darker 1.5.0 (see #279).

.github/workflows/lint_python.yml
akaihola: > error: --non-interactive is only supported with --install-types

Thanks, corrected.

akaihola:
As a general note, if we introduce the additional checks, it would be a bit arbitrary to split the checks between [`.github/workflows/python-package.yml`](https://github.com/akaihola/darker/blob/8a1835c952b9743ea5cb6d40d23eb9c6458e7828/.github/workflows/python-package.yml) and [`.github/workflows/lint_python.yml`](https://github.com/akaihola/darker/pull/310/files#diff-6f698e1e0a7be999d5ea3e4c42773654f2d2f6a0f23b11751c7e0321701ce908). I would rather combine all of them in one workflow. Wouldn't this also be ecological since there wouldn't be multiple workflows setting up the test environment in parallel?

.github/workflows/lint_python.yml
```diff
@@ -7,10 +7,11 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v2
       - run: pip install --upgrade pip wheel
-      - run: pip install bandit codespell darker flake8 flake8-2020 flake8-bugbear
+      - run: pip install bandit black codespell darker flake8 flake8-2020 flake8-bugbear
                          flake8-comprehensions isort mypy pytest pyupgrade safety
       - run: bandit --recursive --skip B101 . || true  # B101 is assert statements
-      - run: darker --check .
+      - run: black --check .
```
akaihola: We intentionally have some non-Black formatting in the code base. This serves as dogfooding so when we run Darker on its own code base, it will skip reformatting those regions unless we've made modifications to them.

So it has been a conscious choice to *not* run Black on Darker's code base.

.github/workflows/lint_python.yml
```diff
@@ -7,10 +7,11 @@ jobs:
...
+      - run: darker --check . || true  # Fails in a GitHub Action
```
akaihola: The workflow should fail if there is any non-Black formatting on modified lines. So ` || true` would need to be removed. But as said earlier, `pytest-darker` already runs `darker` in [.github/workflows/python-package.yml](https://github.com/akaihola/darker/blob/8a1835c952b9743ea5cb6d40d23eb9c6458e7828/.github/workflows/python-package.yml).

.github/workflows/lint_python.yml
```diff
@@ -7,20 +7,15 @@ jobs:
...
-      - run: flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
-      - run: flake8 . --count --exit-zero --max-complexity=10 --max-line-length=88
-                      --show-source --statistics
-      - run: isort --check-only --profile black . || true
+      - run: pip install bandit codespell darker-pytest flake8 flake8-2020
+                         flake8-bugbear flake8-comprehensions mypy pyupgrade safety
+      - run: bandit --recursive  --skip B101,B307,B404,B603 .
+      - run: black --check . || true  # Fails
+      - run: darker --check . || true  # Fails
+      - run: codespell --ignore-words-list="nd,unparseable"
```
akaihola, line 15: Both "parsable" and "parseable" seem to be correct spellings. I think we can follow `codespell`'s advice and correct the spelling to "non-parsable" and "unparsable" in `linting.py`. So that could be left out of the ignore list.

.github/workflows/lint_python.yml
```diff
@@ -0,0 +1,19 @@
...
+  lint_python:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+      - run: pip install --upgrade pip wheel
+      - run: pip install bandit codespell flake8 flake8-2020
+                         flake8-bugbear flake8-comprehensions
+                         mypy pyupgrade safety
+      - run: bandit --recursive  --skip B101,B307,B404,B603 .
```
akaihola, line 14: Would it make sense to not skip any checks globally and instead [mark the failing lines with `# nosec`](https://bandit.readthedocs.io/en/latest/config.html#suppressing-individual-lines) when we know we've audited and accepted the check violation?

.github/workflows/lint_python.yml
akaihola: I wrote:
>I agree it might be a better idea to run Darker explicitly instead of through [pytest-darker](https://pypi.org/project/pytest-darker/)

On a second thought, currently the Darker repository also serves as a "test case" for [pytest-darker](https://pypi.org/project/pytest-darker/). Removing the use of `pytest-darker` would
1. prevent us from seeing when we break compatibility between Darker and `pytest-darker`
2. fail to promote `pytest-darker` for the curious eyes who take a look at Darker's workflows

.github/workflows/lint_python.yml
cclauss: Codespell often suggests American spellings in place of British spellings so often I cut some slack.

cclauss:
# OK...

I think you see my approach and can fine-tune things from here.

akaihola:
>I think you see my approach and can fine-tune things from here.

Yes, I think I'll be able to do that. Thanks @cclauss for your effort and excellent ideas!

.github/workflows/lint_python.yml
akaihola: I replaced the `eval` and addressed also all the other remarks from `bandit`.

.github/workflows/lint_python.yml
akaihola: I added `# nosec` to individual lines.

akaihola (approved):
Tests pass, implementation now fits well into existing workflows. Thanks @cclauss for your initiative and effort!

PR #311, opened 2022-02-25
GitHub Action to run pre-commit on all pull requests

cclauss:
This ensures that if a contributor fails to install pre-commit locally, CI will still highlight any issues in their submissions.


akaihola:
This makes most sense if we move all checks from the workflow to pre-commit, and just run pre-commit from the workflow, right?

I'm a bit torn on this, since I also like GitHub Actions showing each different check as a separate item in the UI. Is there any way to make that still happen when doing the checks within pre-commit?

cclauss:
You understand well -- I too am torn about how to implement this but I tend to put the ___fastest___ checks in pre-commit and the slower checks in a separate GitHub Action.

https://pre-commit.ci is supercool because it allows for automatic corrections to be committed back to the PR but that is a whole other CI tool to consider.

akaihola:
[`pre-commit` documentation](https://pre-commit.com/#pre-commit-run) describes the `pre-commit run` option `[hook-id]`:
>Options:
>
>- `[hook-id]`: specify a single hook-id to run only that hook.

>Some example useful invocations:
>- [...]
>- `pre-commit run flake8`: run the `flake8` hook against all staged files.

Using that option, we could still run each check as their own step in the GitHub Workflow, while in developers' environments `git-commit` would run them all.

That would of course require remembering to add new checks also to the workflow after adding them to the pre-commit configuration.

cclauss:
I find the pre-commit output to be quite readable and have never had any contributor complaints about not understanding that output.

akaihola:
For the 1.5.0, we'll keep running checks directly from GitHub workflows. Let's consider `pre-commit` in 1.6.0. Thanks again @cclauss!

akaihola:
I've been experimenting with some GitHub actions which annotate the `Files changed` view of the pull request with messages from linters and checkers.

To combine that with a good pre-commit setup for local use, it seems to me that at least some configuration needs to be duplicated, which is a bit of a bummer.

PR #338, opened 2022-03-20
Build and validate source and wheel distributions

### LINKED ISSUE #317: In a GitHub Workflow, create an sdist and a wheel, validate with `twine check`
Releasing is currently a manual process.

Optimally, GitHub Actions would make and upload releases. But as a first step towards that, it could at least build the `.tar.gz` sdist and a wheel, and validate with `twine check` that they look good.

Also, for tests, the Darker package could be installed from the built wheel instead of by using `pip install --editable`. This would add further guarantees that packaging works correctly.

akaihola:
Fixes #317.

This won't yet store or upload the distributions anywhere. But it makes sure building them succeeds, and that tests work correctly if Darker is installed from the built wheel.


akaihola:
The build failure will be fixed after merging #340 and rebasing this on `master`.

roniemartinez (approved):

PR #340, opened 2022-03-21
Avoid memory leak from using `@lru_cache` on a method

akaihola:
This was an interesting find reported by the [flake8-bugbear](https://pypi.org/project/flake8-bugbear/) plugin:

>```
>B019 Use of `functools.lru_cache` or `functools.cache` on class methods can lead to memory leaks.
>      The cache may retain instance references, preventing garbage collection.
>```

See [Python functools lru_cache with instance methods: release object](https://stackoverflow.com/questions/33672412/python-functools-lru-cache-with-instance-methods-release-object) on StackOverflow for more information.

### DIFF
--- a/src/darker/git.py
+++ b/src/darker/git.py
@@ -351,7 +351,7 @@ def _revision_vs_lines(
 
     The actual implementation is here instead of in
     `EditedLinenumsDiffer._revision_vs_lines` so it is accessibe both as a method and as
-    a module gloabl for use in the `_compare_revisions` function.
+    a module global for use in the `_compare_revisions` function.
 
     :param path_in_repo: Path of the file to compare, relative to repository root
     :param content: The contents to compare to, e.g. from current working tree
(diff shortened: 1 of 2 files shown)


src/darker/git.py
```diff
@@ -344,36 +344,97 @@ def git_get_modified_python_files(
...
+) -> List[int]:
+    """For file `path_in_repo`, return changed line numbers from given revision
+
+    The actual implementation is here instead of in
+    `EditedLinenumsDiffer._revision_vs_lines` so it is accessibe both as a method and as
+    a module global for use in the `_compare_revisions` function.
+
+    :param path_in_repo: Path of the file to compare, relative to repository root
+    :param content: The contents to compare to, e.g. from current working tree
+    :param context_lines: The number of lines to include before and after a change
```
roniemartinez, line 358: `root` and `rev1` are missing

roniemartinez (approved):
LGTM

PR #342, opened 2022-03-22
Handle files not encoded with UTF-8

gdiscry:
Black can format files encoded with an encoding other than UTF-8, based on the encoding cookie present in the first two lines. Darker is also supposed to format those files, but an exception is raised instead.

The exception happens because all the output from git is decoded with UTF-8, including calls to `git show <revision>:<path>` that returns the binary content of a file from the history. If a file used an encoding other than UTF-8, the decoding fails.

I introduce `TextDocument.from_bytes()` to create a `TextDocument` from the raw binary content of a file, automatically detecting the encoding used, and use this new factory when loading a file from disk or from a revision in the git history.

### DIFF
--- a/src/darker/git.py
+++ b/src/darker/git.py
@@ -9,7 +9,7 @@ from datetime import datetime
 from functools import lru_cache
 from pathlib import Path
 from subprocess import DEVNULL, PIPE, CalledProcessError, check_output, run  # nosec
-from typing import Dict, Iterable, List, Set, Tuple
+from typing import Dict, Iterable, List, Optional, Set, Tuple, Union, overload
 
 from darker.diff import diff_and_get_opcodes, opcodes_to_edit_linenums
 from darker.utils import GIT_DATEFORMAT, TextDocument
@@ -79,8 +79,8 @@ def git_get_content_at_revision(path: Path, revision: str, cwd: Path) -> TextDoc
         return TextDocument.from_file(abspath)
     cmd = ["show", f"{revision}:./{path.as_posix()}"]
     try:
-        return TextDocument.from_lines(
-            _git_check_output_lines(cmd, cwd, exit_on_error=False),
+        return TextDocument.from_bytes(
+            _git_check_output(cmd, cwd, exit_on_error=False),
             mtime=git_get_mtime_at_commit(path, revision, cwd),
         )
     except CalledProcessError as exc_info:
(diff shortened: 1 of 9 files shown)


akaihola:
Hi @gdiscry,

Thanks for the excellent pull request!

There seem to be [a few checks failing](https://github.com/akaihola/darker/runs/5705510112?check_suite_focus=true) due to unused imports, missing docstrings and such. I'll also take a closer look at the patch soon.

src/darker/git.py
```diff
@@ -207,15 +208,49 @@ def _git_check_output_lines(
...
+@overload
+def _git_check_output(
+    cmd: List[str], cwd: Path, exit_on_error: bool, encoding: str
+) -> str:
+    ...
+
+
+@overload
+def _git_check_output(cmd: List[str], cwd: Path, *, encoding: str) -> str:
+    ...
```
akaihola, line 235: I guess there's no way around having to define these both, instead of just one `-> str` variant wtih a default value for `exit_on_error`?

src/darker/tests/test_git.py
```diff
@@ -403,6 +403,30 @@ def test_git_get_content_at_revision_stderr(git_repo, capfd, caplog):
...
+        content = ("\n".join(lines) + "\n").encode(encoding)
+        git_repo.add({"file.py": content}, commit=f"{encoding}")
+        commit = git_repo.get_hash()[:7]
+        commits.append((commit, encoding, lines))
+    commits.append((":WORKTREE:", encoding, lines))
+
+    for commit, encoding, lines in commits:
+        result = git.git_get_content_at_revision(Path("file.py"), commit, git_repo.root)
+        assert result.lines == lines
+        assert result.encoding == encoding
```
akaihola, line 427: If the test for one of the commits fails, the remaining commits won't be tested at all. That's a bit contrary to unit test best practices.

Maybe populate the repository in a fixture, and turn the `for` loop assertions into a parameterized test?

Or, we could take [pytest-assume](https://github.com/astraw38/pytest-assume) into use.

src/darker/tests/test_utils.py
```diff
@@ -229,29 +229,35 @@ def test_textdocument_from_str(
...
+    """TextDocument.from_file()/.from_bytes() detects the file encoding correctly"""
     path = tmp_path / "test.py"
     path.write_bytes(content)
 
     textdocument = TextDocument.from_file(path)
 
     assert textdocument.encoding == expect
 
+    textdocument = TextDocument.from_bytes(content)
+    assert textdocument.encoding == expect
```
akaihola, line 242: In these tests, if `from_file()` fails, `from_bytes` won't be tested at all. Again, that's not optimal for a unit test.

This could be improved e.g. by doing another parameterization by the `.from_*()` method name, or by using [pytest-assume](https://github.com/astraw38/pytest-assume).

akaihola (changes requested):
I made a few review comments. These are minor things, and I'm open to be convinced about your original version.

gdiscry (commented):
There's no need to convince you, your comments match some issues I already had. A new version is coming that addresses your comments.

akaihola:
Excellent! Thanks again for an outstanding pull request.

Mypy seems to be still a bit [unhappy](https://github.com/akaihola/darker/runs/5725924977?check_suite_focus=true) about a few things. Those should be easy to fix with some docstrings and ignores. Let me know if you'd prefer me to do those:
```
src/darker/tests/test_utils.py:21:0: C0116: Missing function or method docstring (missing-function-docstring)
src/darker/tests/test_utils.py:21:0: R1710: Either all return statements in a function should return an expression, or none of them should. (inconsistent-return-statements)

src/darker/tests/test_utils.py:35:26: W0621: Redefining name 'textdocument_factory' from outer scope (line 21) (redefined-outer-name)
src/darker/tests/test_utils.py:35:0: C0116: Missing function or method docstring (missing-function-docstring)

src/darker/tests/test_utils.py:252:38: W0621: Redefining name 'textdocument' from outer scope (line 35) (redefined-outer-name)

src/darker/tests/test_utils.py:262:37: W0621: Redefining name 'textdocument' from outer scope (line 35) (redefined-outer-name)
```

gdiscry:
It's OK, I found out how to run the tests locally so I fixed remaining issues 🤞 (and slightly reworked the `textdocument_factory` fixture).

akaihola (approved):
Perfect!

PR #313, opened 2022-02-25
GitHub Action options to invoke linters through Darker

akaihola:
Fixes #281.

Follow-up issue #337 will allow pinning linter packages.

PR #344, opened 2022-03-24
Fix bandit reporting in GitHub Workflows by using my own branch of the action

akaihola:
In the [akaihola/bandit-report-artifacts@use-config](https://github.com/akaihola/bandit-report-artifacts/commits/use-config) branch (see ioggstream/bandit-report-artifacts#8), the second `bandit` call from inside `main.py` now uses the configuration file as well. Annotations should now match the check status when looking at GitHub Workflow results.

### DIFF
--- a/.github/workflows/bandit.yml
+++ b/.github/workflows/bandit.yml
@@ -11,13 +11,13 @@ jobs:
       - uses: actions/checkout@v2
 
       - name: Bandit security check for code
-        uses: ioggstream/bandit-report-artifacts@v0.0.2
+        uses: akaihola/bandit-report-artifacts@use-config
         with:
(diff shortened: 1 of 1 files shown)


akaihola:
FYI @ioggstream, the fix in ioggstream/bandit-report-artifacts#8 seems to work. The `bandit` checks for test directories in this project no longer ignore the configuration, so `B101` is actually disabled and GitHub Workflows don't display unwanted annotations nor do they fail the build now.

See the screenshot in #313 or what the problem looked like before this fix.

Also ping @cclauss, a `bandit` problem was fixed here.

roniemartinez (approved):
LGTM

PR #349, opened 2022-03-30
Allow constraining linter versions in GitHub Action configuration

### LINKED ISSUE #337: Allow pinning linter packages in the GitHub Action
To keep GitHub workflows stable, users may want to pin both Darker and linter packages to fixed versions when using the GitHub Action. Darker can already be pinned using the `version:` option, but linter pinning wasn't yet added in #313.

# Alternatives for the syntax

In these examples, Pylint is pinned but the newest version of Flake8 is to be installed.

## Embed versions in `lint:` option
In this alternative, items in `lint:` need to be validated and passed to `pip install`, and command names without the version are needed for the Darker command line `--lint` options. For linters with no version number, the newest version on PyPI would be installed:

`.github/workflows/darker.yml`:
```yaml
      - uses: akaihola/darker@1.5.0
        with:
          lint: pylint==2.12.2, flake8
```

## A separate `constraints:` option
Here, versions of packages are provided in a separate `constraints:` option. If a linter is omitted from `constraints`, the newest version would be installed.

The use of this option could extend beyond just linters. We could even deprecate the `version:` option and guide users to use `constraints: darker==1.5.0` instead. And this could also solve the missing `isort` version pinning feature.

`.github/workflows/darker.yml`:
```yaml
      - uses: akaihola/darker@1.5.0
        with:
          lint: pylint, flake8
          constraints: pylint==2.12.2
```

## Constraints in a text file
We could also require the user to provide constraints in a text file and point to it in a `constraints-file:` option:

`.github/workflows/darker.yml`:
```yaml
      - uses: akaihola/darker@1.5.0
        with:
          lint: pylint, flake8
          constraints-file: .github/workflows/darker-lint-constraints.txt
```

`.github/workflows/darker-lint-contstraints.txt`:
```
pylint==2.12.2
```

# Considerations

- Is the Pip constraints feature well enough known, or would it be better to use the name `versions:` instead of `constraints:`?

akaihola:
Fixes #337. Being tested in akaihola/pgtricks#18.

Linter support was added in #313.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,7 +7,7 @@ Added
 -----
 - The ``--jobs`` option now specifies how many Darker jobs are used to process files in
   parallel to complete reformatting/linting faster.
-- Linters can now be run in the GitHub Action using the ``lint:`` option.
+- Linters can now be installed and run in the GitHub Action using the ``lint:`` option.
   
 Fixed
 -----
--- a/README.rst
+++ b/README.rst
@@ -575,7 +575,7 @@ Create a file named ``.github/workflows/darker.yml`` inside your repository with
              revision: "master..."
              src: "./src"
              version: "1.4.2"
-             lint: "flake8,pylint"
+             lint: "flake8,pylint==2.13.1"
 
 There needs to be a working Python environment, set up using ``actions/setup-python``
 in the above example. Darker will be installed in an isolated virtualenv to prevent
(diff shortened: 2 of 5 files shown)


akaihola:
The workflow in akaihola/pgtricks#18 seems to work. I have this in `.github/workflows/python-package.yml`:
```yaml
    - uses: akaihola/darker@gh-action-linter-versions
      with:
        options: --verbose --check --diff --isort
        revision: origin/master...
        lint: flake8 == 4.0.0 , pylint==2.13.2
```

The versions for `flake8` and `pylint` are one patch version older than the newest releases. And sure enough, I see this in the log for the workflow:
```
Successfully installed astroid-2.11.2 black-22.3.0 click-8.1.0 darker-1.4.2 dill-0.3.4 flake8-4.0.0
isort-5.10.1 lazy-object-proxy-1.7.1 mccabe-0.6.1 mypy-extensions-0.4.3 pathspec-0.9.0 platformdirs-2.5.1
pycodestyle-2.8.0 pyflakes-2.4.0 pylint-2.13.2 toml-0.10.2 tomli-2.0.1 wrapt-1.14.0
```
So `flake8==4.0.0` and `pylint==2.13.2` were indeed installed.

Note that the workflow fails on purpose – I added extra spaces so reformatters and linters would fail and show the log.

roniemartinez (approved):
Awesome! 🚀

PR #352, opened 2022-04-01
Rename `--jobs` to `-W` / `--workers`

### LINKED ISSUE #335: `--jobs` option name doesn't match the corresponding `-W`/`--workers` in Black
We've recently merged #179 for inclusion in the [1.5.0 release](https://github.com/akaihola/darker/milestone/13). It adds a `--jobs` argument. However, I only just noticed Black has a similar feature and the command line option is
```
-W, --workers INTEGER RANGE     Number of parallel workers  [default: 8; x>=1]
```

Should we adopt the same option instead of `--jobs` and do this rename before the 1.5.0 release?

This might avoid difficulties when using Darker as a drop-in replacement for Black e.g. in IDEs.

I think it also makes sense to imitate Black's user interface as much as possible just because it's the closest tool by purpose to Darker, and because many users are probably migrating to Darker or taking it into use after using or trying out Black.

akaihola:
Fixes #335.

The `--jobs` option was originally proposed in #178 and implemented in #179.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,8 +5,8 @@ These features will be included in the next release:
 
 Added
 -----
-- The ``--jobs`` option now specifies how many Darker jobs are used to process files in
-  parallel to complete reformatting/linting faster.
+- The ``--workers``/``-W`` option now specifies how many Darker jobs are used to
+  process files in parallel to complete reformatting/linting faster.
 - Linters can now be installed and run in the GitHub Action using the ``lint:`` option.
   
 Fixed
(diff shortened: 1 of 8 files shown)


akaihola:
@rogalski, would you like to review this option rename in Darker?

akaihola:
@MatthijsBurgh & @roniemartinez: as the original authors of the `--jobs` feature seem to be busy and unable to review this, may I ask you whether you'd agree on changing the name of the option from `--jobs` to `-W`/`--workers` (similar to Black)? Are there some viewpoints which would support `--jobs` as a better name?

roniemartinez (approved):
LGTM

akaihola:
Thanks @roniemartinez!

Issue #24, opened 2020-07-09
Read black settings from pyproject.toml


DylanYoung:
Just banging my head against this with vscode, so thought I'd check the source.  

Correct me if I'm wrong but pyproject.toml isn't currently supported here, right?

It'd be a nice enhancement if it were, but if not, it'd be nice to note in the docs that it doesn't work.

akaihola:
Support for `pyproject.toml` is coming in the next release. You can try it out by installing `darker` from the `master` branch.

DylanYoung:
I wasn't sure if that included the auto-detection or just the command line flags.   I thought I still had to pass the flags in my editor, but perhaps I fudged up my testing.

DylanYoung:
I can see that the unittest is checking this case though (looks like it is anyways).  Thanks @akaihola!  I wonder if an update to the changelog to mention that auto-detection is working as well is worthwhile?   If you think so, I can modify this issue to reflect that instead.   Otherwise, you can close it :)

akaihola:
@DylanYoung I'm all for mentioning the auto-detection in the changelog, go ahead!

Also, I believe it isn't described in the documentation either. That could still be addressed in #22.

akaihola:
@DylanYoung, do the additions to the `README` in 422996b match with what you had in mind?

DylanYoung:
Absolutely! Closed by 422996b and PR #1

PR #353, opened 2022-04-03
Allow user to force syntax highlighting using config file, env var or command line. Enable syntax highlighting in GitHub Action.

### LINKED ISSUE #263: The GitHub action for Darker doesn't use syntax highlighting
Currently, you run:
`proc = subprocess.run([sys.executable, str(MAIN_SCRIPT)]);`
in the github action. 

It doesn't output any colors, which makes it harder to understand darker's output.
You could use something like written here: https://stackoverflow.com/questions/58307157/subprocess-call-output-in-color printing the output with colors.

akaihola:
By default, highlighting is now enabled if outputting to a terminal and `pygments` is installed.

Highlighting can now be force enabled for non-terminal output or force disabled for terminal output using:
- `color = {true|false}` in `pyproject.toml`,
- the `PY_COLORS={0|1}` environment variables (overrides `pyproject.toml`), or
- the `--color`/`--no-color` command line options (override both `pyproject.toml` and the environment variable)

Remaining tasks:
- [x] Document the new configuration options
- [x] Update command line help in documentation
- [x] Change log
- [x] Verify that this works for real in akaihola/pgtricks#18
- [x] Install Pygments in the GitHub Action
- [x] Use `--color` by default in the GitHub Action
- [x] Allow installing Darker from a branch in the GitHub Action (to enable testing of feature branches)
- [ ] Decide whether to include this in [1.5.0](https://github.com/akaihola/darker/milestone/13) with or without #355, or to postpone this for [1.5.1](https://github.com/akaihola/darker/milestone/15) (see [discussion](https://github.com/akaihola/darker/milestone/15))

Fixes #263.

See also #355 as a follow-up issue (possible support for the `NO_COLOR` environment variable).


akaihola:
It works in akaihola/pgtricks#18!

kuva

akaihola:
@mayk0gan, good news, you can see syntax highlighting in action for Darker output in [akaihola/pgtricks build #24](https://github.com/akaihola/pgtricks/runs/5839773231?check_suite_focus=true) (see screenshot in earlier comment).

Could you review this pull request and approve it for merging if it looks good to you?

akaihola:
Looking for a reviewer for this, pinging contributors previously interested in the GitHub Action:
@jenshnielsen @cclauss @brtknr @yoursvivek

README.rst
```diff
@@ -723,6 +736,32 @@ Here's an example of `cov_to_lint.py`_ output::
 .. _cov_to_lint.py: https://gist.github.com/akaihola/2511fe7d2f29f219cb995649afd3d8d2
 
 
+Syntax highlighting
+===================
+
+Darker automatically enables syntax highlighting for the ``--diff``,
+``-d``/``--stdout`` and ``-L``/``--lint`` options if it's run on a terminal and the
```
roniemartinez, line 743:
```suggestion
``-d``/``--stdout`` and ``-L``/``--lint`` options if it's running on a terminal and the
```

src/darker/tests/test_highlighting.py
```diff
@@ -1,50 +1,174 @@
...
-
-        from darker.highlighting import colorize
-
-        assert colorize == without_pygments.colorize
-
-
-def test_colorize_import_with_pygments():
-    """The real ``colorize()`` is used if Pygments is available"""
-    assert "pygments" in sys.modules
+@pytest.mark.parametrize(
```
roniemartinez, line 15: [Suggestion] You can use 2 or more parametrize decorator to create a matrix here instead of manually inputting the params.

roniemartinez (approved):
LGTM

src/darker/tests/test_highlighting.py
akaihola: @roniemartinez thanks for the suggestion!

I actually prefer the explicit, human-readable list of parameter combinations and expected behavior. This way it's easy to read and understand what the behavior is and should be.

Also, the expected value (`USE_COLOR` or not) depends on the input values in a non-trivial way. If the test function derived it from values of parameters in the matrix, it would need to re-implement the function being tested.

In general, I've grown to appreciate the principle that DRY ("don't repeat yourself") doesn't apply to test code – instead test code should strive for maximum simplicity, clarity and understandability. Otherwise there's a risk that there's a bug not only in the code being tested but also the complicated test code.

In this particular case, it would be nice if the initialization of environment variables, `pyproject.toml`, command line options and the availability of the `pygments` module was simpler while still preserving the readability of input values and the expected value, but I couldn't find a better solution than this.

akaihola: Actually, since `should_use_color()` returns `False` always if `pygments` can't be imported, that half of the test cases _could_ actually be separated into its own test case with matrix parameterization. I'll experiment with that.

src/darker/config.py
```diff
@@ -97,6 +99,21 @@ def validate_config_output_mode(config: DarkerConfig) -> None:
...
 
+def override_color_with_environment(pyproject_config: DarkerConfig) -> DarkerConfig:
+    """Override ``color`` if the ``PY_COLORS`` environment variable is '0' or '1'
+
+    :param config: The configuration read from ``pyproject.toml``
+    :return: The modified configuration
+
+    """
+    py_colors = os.getenv("PY_COLORS")
+    if py_colors not in {"0", "1"}:
```
MatthijsBurgh, line 110: I think `true` or `false` should also be accepted. (Including all capital case variants)

src/darker/help.py
```diff
@@ -53,13 +55,23 @@
...
 VERBOSE = "Show steps taken and summarize modifications"
 QUIET = "Reduce amount of output"
+COLOR = (
+    "Enable syntax highlighting even for non-terminal output. Overrides the environment"
+    " variable PY_COLORS=0"
+)
+NO_COLOR = (
+    "Disable syntax highlighting even for terminal output. Overrides the environment"
+    " variable PY_COLORS=1"
+)
```
MatthijsBurgh, line 74: When allowing `true` and `false` for colors, this should be changed too.

src/darker/highlighting/__init__.py
```diff
@@ -1,12 +1,54 @@
...
+    else:
+        use_color = sys.stdout.isatty()
+    if use_color:
+        try:
+            import pygments  # noqa
+
+            return True
+        except ImportError:
+            pass
+    return False
```
MatthijsBurgh, line 38:
```suggestion
        except ImportError:
            return False
```

src/darker/tests/test_command_line.py
```diff
@@ -241,6 +242,69 @@ def test_parse_command_line_config_src(
...
+        expect_config=("color", False),
+        expect_modified=("color", False),
+    ),
+    dict(
+        argv=["--no-color", "."],
+        environ={"PY_COLORS": "1"},
+        expect_value=("color", False),
+        expect_config=("color", False),
+        expect_modified=("color", False),
+    ),
```
MatthijsBurgh, line 307: When allowing `true` and `false` for colors, this should be extended.

src/darker/tests/test_highlighting.py
MatthijsBurgh: I favor the suggestion of @roniemartinez. A human-readable list would be preffered, when it is a short list. This is a very long list and therefore not beneficial IMO.

src/darker/tests/test_highlighting.py
```diff
@@ -1,50 +1,174 @@
...
+        print("[tool.darker]", file=pyproject)
+        if " color=True " in params:
+            print("color = true", file=pyproject)
+        if " color=False " in params:
+            print("color = false", file=pyproject)
+    env = {}
+    if " PY_COLORS=0 " in params:
+        env["PY_COLORS"] = "0"
+    if " PY_COLORS=1 " in params:
+        env["PY_COLORS"] = "1"
```
MatthijsBurgh, line 145: This should be extend when allowing `true` and `false` for color.

MatthijsBurgh (changes requested):
I think we should allow `true` and `false` for color.

src/darker/tests/test_highlighting.py
akaihola: @roniemartinez, I now converted the non-`pygments` half of these test cases into a parameterized matrix test. I also touched the remaining half a bit to hopefully make it even more understandable but also more concise.

What do you think?

src/darker/tests/test_highlighting.py
akaihola: Ah, thanks @MatthijsBurgh, I didn't think of digging further to see whether other values than `0` and `1` are generally accepted for `PY_COLORS` by other Python tools. I'll double check – or do you have such examples readily in your mind?

akaihola: Based on a quick search on GitHub for [`PY_COLORS true`](https://github.com/search?p=2&q=PY_COLORS+true&type=Code) and [`PY_COLORS false`](https://github.com/search?p=2&q=PY_COLORS+false&type=Code), the only instance of non-`0`/`1` values is [here in Ansible `molecule`](https://github.com/ansible-community/molecule/blob/e78c820c6efdf09eb14432d15fca546730a1b37a/src/molecule/console.py#L45-L50):

It uses the [`to_bool()`](https://github.com/ansible-community/molecule/blob/e78c820c6efdf09eb14432d15fca546730a1b37a/src/molecule/console.py#L30-L38) helper, which has:
```python
    if isinstance(a, str):
        a = a.lower()
    if a in ("yes", "on", "1", "true", 1):
        return True
```

But since this is done not only when the `PY_COLORS` environment variable is encountered, but also `CLICOLOR`, `FORCE_COLOR` and `ANSIBLE_FORCE_COLOR`, it's not clear from which ones of these other values than `0` and `1` are to be expected.

Pytest documentation [says](https://docs.pytest.org/en/6.2.x/reference.html#envvar-PY_COLORS):
>**PY_COLORS**
>When set to `1`, pytest will use color in terminal output. When set to `0`, pytest will not use color. `PY_COLORS` takes precedence over `NO_COLOR` and `FORCE_COLOR`.

Textualize/rich#343 has some discussion on `PY_COLORS`, and only mentions `0`/`1`.

Many packages seem to rely on `py.io.TerminalWriter` to handle color output. Support for `PY_COLORS` seems to have been introduced in version 1.4.19 and is mentioned in the [change log](https://py.readthedocs.io/en/latest/changelog.html?highlight=PY_COLORS#changes-between-1-4-18-and-1-4-19):
>- check `PY_COLORS=1` or `PY_COLORS=0` to force coloring/not-coloring for `py.io.TerminalWriter()` independently from capabilities of the output file. Thanks Marc Abramowitz for the PR.

Based on the above evidence, I don't think we should jump into supporting other values than `PY_COLORS=0` and `PY_COLORS=1`. Would you agree?

src/darker/tests/test_highlighting.py
akaihola: @MatthijsBurgh, how would you implement it so that the rules for determining the expected `True`/`False` value don't just add a second (and similarly bug-risky) implementation of the algorithm which is being tested?

Also, what do you think about the length of the list of test cases after I moved half of them into a parameterized matrix test?

akaihola: I guess two thirds of the remaining test cases could be extracted into two separate test functions for the cases with the `--color` and `--no-color` command-line options, which always force highlighting on or off, respectively, if `pygments` is installed.

But now, instead of one long easily eye-scannable list of test cases, we'd have four different test functions, increasing the chance of mistakes in the test code, and making it more challenging to read and understand the tests when debugging possible future failures.

So I'm not 100% sure this is a good idea either. Please convince me otherwise!

src/darker/tests/test_highlighting.py
MatthijsBurgh: Haha, Good research. Then we should stick to `0/1`.

akaihola: pypa/pip#10909 is so far the "highest profile" discussion I've found on this topic.

It seems that [pip](https://github.com/pypa/pip) and [colorama](https://github.com/tartley/colorama) are going to support only `FORCE_COLOR=<anything>` and `NO_COLOR=<anything>`, but not `PY_COLORS`.

On the other hand, Pytest does support `PY_COLORS={1|0}` and `NO_COLOR=<anything>` (with `PY_COLORS` taking precedence).

So in the end it may be best to do the same as Pytest:
- `PY_COLORS={1|0}` (taking precedence over other options)
- `NO_COLOR=<anything>`
- and possibly `FORCE_COLOR=<anything>` (similar to [pip](https://github.com/pypa/pip) and [colorama](https://github.com/tartley/colorama))

As for the `true`/`false` discussion, I'm going to make the call to only support `PY_COLORS={0|1}` for now. Thanks @MatthijsBurgh for initiating the discussion and motivating me to look further into the topic!

src/darker/highlighting/__init__.py
akaihola: Darker will only support `PY_COLORS=0` and `PY_COLORS=1`, see [discussion here]. Thanks @MatthijsBurgh for the suggestion!

src/darker/tests/test_highlighting.py
akaihola: I added #355 for adding `NO_COLOR` (and possibly `FORCE_COLOR`) support in the future.

src/darker/highlighting/__init__.py
MatthijsBurgh: I see that my suggestion has a different behaviour. I missed the last `if`.

src/darker/highlighting/__init__.py
```diff
@@ -1,12 +1,54 @@
...
+                         overridden using environment variables and/or command line
+                         options
+    :return: ``True`` if highlighting should be used
+
+    """
+    if config_color is not None:
+        use_color = config_color
+    else:
+        use_color = sys.stdout.isatty()
+    if use_color:
```
MatthijsBurgh, line 31:
```suggestion
    if config_color is not None:
        use_color = config_color
    else:
        use_color = sys.stdout.isatty()

    if use_color:
```

Keep an empty line in between different `if/else` blocks

src/darker/tests/test_highlighting.py
MatthijsBurgh: Does pytest show the chosen values for failing runs? If that is the case, I think we should go for the matrix.

> But now, instead of one long easily eye-scannable list of test cases, we'd have four different test functions

Do you mean four different functions or decorators?

roniemartinez: @akaihola Just to add clarity into the concept of the parametrize matrix. I think this is more readable.

```python
@pytest.mark.parametrize("tty", ["tty", ""])
@pytest.mark.parametrize("arg", ["--color", ""])
@pytest.mark.parametrize("env", ["1", "0"])
@pytest.mark.parametrize("color", [True, False])
def test_parametrized_matrix(color, env, arg, tty):
    assert True
``` 

This in turn generates 16 tests

```
collected 16 items                                                                                                               

tests/test_sample.py::test_parametrized_matrix[True-1---color-tty] PASSED                                                  [  6%]
tests/test_sample.py::test_parametrized_matrix[True-1---color-] PASSED                                                     [ 12%]
tests/test_sample.py::test_parametrized_matrix[True-1--tty] PASSED                                                         [ 18%]
tests/test_sample.py::test_parametrized_matrix[True-1--] PASSED                                                            [ 25%]
tests/test_sample.py::test_parametrized_matrix[True-0---color-tty] PASSED                                                  [ 31%]
tests/test_sample.py::test_parametrized_matrix[True-0---color-] PASSED                                                     [ 37%]
tests/test_sample.py::test_parametrized_matrix[True-0--tty] PASSED                                                         [ 43%]
tests/test_sample.py::test_parametrized_matrix[True-0--] PASSED                                                            [ 50%]
tests/test_sample.py::test_parametrized_matrix[False-1---color-tty] PASSED                                                 [ 56%]
tests/test_sample.py::test_parametrized_matrix[False-1---color-] PASSED                                                    [ 62%]
tests/test_sample.py::test_parametrized_matrix[False-1--tty] PASSED                                                        [ 68%]
tests/test_sample.py::test_parametrized_matrix[False-1--] PASSED                                                           [ 75%]
tests/test_sample.py::test_parametrized_matrix[False-0---color-tty] PASSED                                                 [ 81%]
tests/test_sample.py::test_parametrized_matrix[False-0---color-] PASSED                                                    [ 87%]
tests/test_sample.py::test_parametrized_matrix[False-0--tty] PASSED                                                        [ 93%]
tests/test_sample.py::test_parametrized_matrix[False-0--] PASSED                                                           [100%]
```

akaihola: Pytest does show the chosen values for failing runs.

I meant four different test functions if I split out the "simple" cases:
- `pygments` not installed -> always `False`
- `pygments` installed and `--color` used -> always `True`
- `pygments` installed and `--no-color` used -> always `False`
- `pygments` installed and no command line option -> depends on `pyproject.toml` and environment variables; This would be either a parametrized list of test cases or a matrix.

But if we go with a matrix either for that 4th set of remaining test cases, or for the whole shebang, my question remains:
>How would you implement the test so that the rules for determining the expected `True`/`False` value don't just add a second (and similarly bug-risky) implementation of the algorithm which is being tested (i.e. the logic in `should_use_color()`)?

akaihola: @roniemartinez thanks for the clear code example.

However, you haven't implemented the assertion yet in the example, and that's precisely the tricky part. The expected return value from `should_use_color()` _depends on the input values in a complex way_.

How do you handle that in a matrix parameterization without re-implementing `should_use_color()`'s logic in the test function?

(this is essentially the same question as in me previous comment, reworded)

roniemartinez: I see `--no-color` forces everything to not use color. I would create a separate test for that to simplify it. As for the other conditions, they seem to be just an "OR" condition. I would say 2 tests with matrices. Please correct the inaccuracy.

```python
# None means not present


@pytest.mark.parametrize("tty", ["tty", None])
@pytest.mark.parametrize("env", [1, 0, None])
@pytest.mark.parametrize("color", [True, False, None])
def test_not_use_color(color, env, tty):
    # with --no-color should always disable color
    ...
    assert should_use_color() is False

@pytest.mark.parametrize("tty", ["tty", None])
@pytest.mark.parametrize("arg", ["--color", None])
@pytest.mark.parametrize("env", [1, 0, None])
@pytest.mark.parametrize("color", [True, False, None])
def test_matrix(color, env, arg, tty):
    # at this point `color or env or arg or tty` should determine the output
    ...
    assert should_use_color() == color or env or arg or tty
```

akaihola: >`--no-color` forces everything to not use color. I would create a separate test for that to simplify it

Yes, that's probably a decent idea. And it's what I proposed earlier above, and I'm currently experimenting with it in #356 (see commit db57c55). It does lead to repetition of test setup code, and I'll probably refactor it into fixtures.

>other conditions, they seem to be just an "OR" condition

Yes, and I argue that writing `color or env or arg or tty` is just a contracted re-implementation of the combination of `parse_command_line()` and `should_use_color()`. The latter, for example, has:
```python
    if config_color is not None:
        use_color = config_color
    else:
        use_color = sys.stdout.isatty()

    if use_color:
        try:
            import pygments  # noqa

            return True
        except ImportError:
            pass
    return False
```
So if we make an aberration (or a typo, or whatever) in the "OR" clause, that same aberration might exist in this code which we're testing.

I still stand behind my principle of avoiding re-implementation of the logic being tested.

In other words, my rule of thumb would be:

>_A test case should declare the expected value, not calculate it._

akaihola: Anyway, I'm going to refine the tests in #356 since the addition of `NO_COLOR` again doubles the number of test cases and calls for refactoring of the tests.

I'll include both #353 and #356 in [1.5.0](https://github.com/akaihola/darker/milestone/13), so we can continue work on the test suite there.

Big thanks to you both, @roniemartinez and @MatthijsBurgh, for working on this with me!

PR #354, opened 2022-04-06
(action) fix getting revision and commit_range

MatthijsBurgh:
`os.getenv` default is only used if key is not defined, not if the value resolves to False.

### DIFF
--- a/action/main.py
+++ b/action/main.py
@@ -14,9 +14,13 @@ OPTIONS = os.getenv("INPUT_OPTIONS", default="")
 SRC = os.getenv("INPUT_SRC", default="")
 VERSION = os.getenv("INPUT_VERSION", default="")
 LINT = os.getenv("INPUT_LINT", default="")
-REVISION = os.getenv(
-    "INPUT_REVISION", default=os.getenv("INPUT_COMMIT_RANGE", default="HEAD^")
-)
+
+REVISION = os.getenv("INPUT_REVISION")
+COMMIT_RANGE = os.getenv("INPUT_COMMIT_RANGE")
+if not REVISION:
+    if not COMMIT_RANGE:
+        COMMIT_RANGE = "HEAD^"
+    REVISION = COMMIT_RANGE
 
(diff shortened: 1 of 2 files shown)


akaihola:
Is it so that the below block in `action.yml` will result in an empty string value for inputs which don't appear in the `with:` block?
```yaml
      env:
        # TODO: Remove once https://github.com/actions/runner/issues/665 is
        #       fixed.
        INPUT_OPTIONS: ${{ inputs.options }}
        INPUT_SRC: ${{ inputs.src }}
        INPUT_VERSION: ${{ inputs.version }}
        INPUT_REVISION: ${{ inputs.revision }}
        INPUT_LINT: ${{ inputs.lint }}
        INPUT_COMMIT_RANGE: ${{ steps.commit-range.outputs.commit-range }}
        pythonioencoding: utf-8
```
If that's true, then I guess `default=""` is redundant in the other `os.getenv()` calls, too?

action/main.py
```diff
@@ -14,9 +14,13 @@
...
-REVISION = os.getenv(
-    "INPUT_REVISION", default=os.getenv("INPUT_COMMIT_RANGE", default="HEAD^")
-)
+
+REVISION = os.getenv("INPUT_REVISION")
+COMMIT_RANGE = os.getenv("INPUT_COMMIT_RANGE")
+if not REVISION:
+    if not COMMIT_RANGE:
+        COMMIT_RANGE = "HEAD^"
+    REVISION = COMMIT_RANGE
```
akaihola, line 23: `COMMIT_RANGE` isn't used anywhere, so I guess this could be rewritten as e.g.
```suggestion
    REVISION = os.getenv("INPUT_COMMIT_RANGE") or "HEAD^"
```
and also line 19 could be deleted?

I also now thought about all-whitespace values (which wouldn't be False values), but it seems the YAML code `key:<space><space><space>` is decoded as {"key": None}, so that really isn't a concern.

akaihola (approved):

akaihola:
Thanks @MatthijsBurgh!

PR #356, opened 2022-04-10
Support the `NO_COLOR` environment variable

### LINKED ISSUE #355: Add support for the `NO_COLOR` environment variable
pypa/pip#10909 is so far the "highest profile" discussion I've found on this topic.

It seems that [pip](https://github.com/pypa/pip) and [colorama](https://github.com/tartley/colorama) are going to support only `FORCE_COLOR=<anything>` and `NO_COLOR=<anything>`, but not `PY_COLORS`.

On the other hand, Pytest does support `PY_COLORS={1|0}` and `NO_COLOR=<anything>` (with `PY_COLORS` taking precedence).

So in the end it may be best to do the same as Pytest:
- `PY_COLORS={1|0}` (taking precedence over other options)
- `NO_COLOR=<anything>`
- and possibly `FORCE_COLOR=<anything>` (similar to [pip](https://github.com/pypa/pip) and [colorama](https://github.com/tartley/colorama))

As for the `true`/`false` discussion, I'm going to make the call to only support `PY_COLORS={0|1}` for now. Thanks @MatthijsBurgh for initiating the discussion and motivating me to look further into the topic!

_Originally posted by @akaihola in https://github.com/akaihola/darker/pull/353#discussion_r846752942_

akaihola:
Setting `NO_COLOR` to any non-empty value will disable syntax highlighting even on terminal output. This overrides the `color = true` setting in `pyproject.toml`, but will be overridden by `PY_COLORS=1` or the `--color` command line option, if those are present.

Fixes #355.

Previously, #353 added a command line options, a `pyproject.toml` option and the `PY_COLORS` environment variable to for enabling/disabling syntax highlighting.

See [earlier discussion](https://github.com/akaihola/darker/pull/353#discussion_r846746769) in #353, and pypa/pip#10909, for the reasoning behind supporting both `NO_COLOR` and `PY_COLORS`.

- [x] decide whether it would make sense to support `FORCE_COLOR` as well (see [this comment](https://github.com/pypa/pip/issues/10909#issuecomment-1066153211) in pypa/pip)
- [x] add suppert for `FORCE_COLOR`
- [x] empty value of `NO_COLOR` and `FORCE_COLOR` should be considered as activating that option, similar to how Pytest works
- [x] tests
- [x] ensure fast performing tests
- [x] change log
- [x] code review

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,7 +11,7 @@ Added
 - Sort imports only if the range of modified lines overlaps with changes resulting from
   sorting the imports.
 - Allow force enabling/disabling of syntax highlighting using the ``color`` option in
-  ``pyproject.toml``, the ``PY_COLORS`` environment variable, and the
+  ``pyproject.toml``, the ``PY_COLORS`` and ``NO_COLOR`` environment variables, and the
   ``--color``/``--no-color`` command line options.
 - Syntax highlighting is now enabled by default in the GitHub Action.
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,7 @@ Added
   ``pyproject.toml``, the ``PY_COLORS`` and ``NO_COLOR`` environment variables, and the
   ``--color``/``--no-color`` command line options.
 - Syntax highlighting is now enabled by default in the GitHub Action.
+- ``pytest>=6.2.0`` now required for the test suite due to type hinting issues.
 
 Fixed
 -----
--- a/constraints-oldest.txt
+++ b/constraints-oldest.txt
@@ -13,7 +13,7 @@ flake8-bugbear==22.1.11
 flake8-comprehensions==3.7.0
 mypy==0.940
 Pygments==2.4.0
-pytest==6.1.0
+pytest==6.2.0
 pytest-flake8==1.0.6
 pytest-isort==1.1.0
 pytest-kwparametrize==0.0.3
--- a/setup.cfg
+++ b/setup.cfg
@@ -62,7 +62,7 @@ test =
     mypy>=0.940
     pygments
     pylint
-    pytest>=6.1.0
+    pytest>=6.2.0
     pytest-darker
     pytest-flake8>=1.0.6
     pytest-isort>=1.1.0
--- a/src/darker/config.py
+++ b/src/darker/config.py
@@ -106,11 +106,12 @@ def override_color_with_environment(pyproject_config: DarkerConfig) -> DarkerCon
     :return: The modified configuration
 
     """
-    py_colors = os.getenv("PY_COLORS")
-    if py_colors not in {"0", "1"}:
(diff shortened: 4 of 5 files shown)


akaihola:
@roniemartinez & @MatthijsBurgh, in addition to adding support for the `NO_COLOR` environment variable, I've now converted the tests for syntax highlighting configuration into parametrized matrix tests as much as possible.

There are now four test functions:
- [`test_should_use_color_no_pygments()`](https://github.com/akaihola/darker/blob/18d5232aa84871746fd6646720fc7eaa8fe666e1/src/darker/tests/test_highlighting.py#L176-L192) verifies that all combinations of configuration do disable syntax highlighting if `pygments` is not present (108 cases)
- [`test_should_use_color_pygments_and_command_line_argument()](https://github.com/akaihola/darker/blob/18d5232aa84871746fd6646720fc7eaa8fe666e1/src/darker/tests/test_highlighting.py#L195-L217)` checks that for all combinations of other configuration options, the command line arguments dictate use of syntax highlighting (72 cases)
- [`test_should_use_color_pygments_and_py_colors()`](https://github.com/akaihola/darker/blob/18d5232aa84871746fd6646720fc7eaa8fe666e1/src/darker/tests/test_highlighting.py#L220-L246) makes sure `PY_COLORS={0|1}` overrides `pyproject.toml` and `NO_COLOR` in all of their combinations (24 cases)
- [`test_should_use_color_pygments()`](https://github.com/akaihola/darker/blob/18d5232aa84871746fd6646720fc7eaa8fe666e1/src/darker/tests/test_highlighting.py#L249-L291) goes through the remaining combinations in which `pyproject.toml`, `NO_COLOR` and tty/non-tty determine use of syntax highlighting in absence of other configuration options (12 cases)

To reduce duplication and boilerplate, I've created [some parametrized fixtures](https://github.com/akaihola/darker/blob/18d5232aa84871746fd6646720fc7eaa8fe666e1/src/darker/tests/test_highlighting.py#L36-L147) instead of straight using `@pytest.mark.parametrize`. This way the patching of environment variables, creating the `pyproject.toml` file and patching the return value of `sys.stdout.isatty` are done inside the fixtures instead of separately in each test function.

I still suspect that for someone fresh looking at the tests, [the resulting 256 lines of Python code](https://github.com/akaihola/darker/blob/18d5232aa84871746fd6646720fc7eaa8fe666e1/src/darker/tests/test_highlighting.py#L36-L291) are more difficult to understand and maintain than the [254 lines (here still just 146 lines since `NO_COLOR` wasn't yet implemented)](https://github.com/akaihola/darker/blob/e78793cdcc034bb4bf1598e4fc81139b31821708/src/darker/tests/test_highlighting.py#L15-L160) it would take by enumerating each parameter combination separately.

roniemartinez (approved):
LGTM

PR #357, opened 2022-04-13
Update `cachix/install-nix-action` to `v17`, downgrade to Python 3.9 on macOS Nix

akaihola:
Fixes macOS build error (see e.g. [`text-nixos (macos-latest)` in build #1248](https://github.com/akaihola/darker/runs/5994916141)).


akaihola:
Still fails on macOS. See cachix/install-nix-action#135. I wonder if we should use Python 3.9 on macOS. On the other hand, this used to work until a couple of days ago.

PR #359, opened 2022-04-17
Use `-c PATH`/`--config PATH` for Darker config

### LINKED ISSUE #244: Darker did rely on `--config` but now only uses black `find_project_root` for it's own config, breaking mono-repository usage
We have a mono-repo layout something like this
```
Monorepo/
├── pyproject.toml
├── lib/
│   ├── some_lib/
│   │   ├── setup.py
│   │   └── pyproject.toml
│   └── another_lib/
│       ├── setup.py
│       └── pyproject.toml
└── apps/
    └── some_app/
        ├── setup.py
        └── pyproject.toml
```

The root `pyproject.toml` contains the `[tool.darker]` and `[tool.black]` configuration for the whole project.
The individual apps/libs pyproject.toml files contain build and other settings.

You used to have the `load_config` for darker respect the `--config` argument: https://github.com/akaihola/darker/blob/dce884b1964b0b883be559e269e39e9b9c329f86/src/darker/config.py#L22

But unfortunately you removed this:
https://github.com/akaihola/darker/blob/4f2e943e42dbf46c1f338022cac9ab89762bc584/src/darker/config.py#L100

Now if we have changed for example `lib/some_lib/setup.py` then the sources that get passed are `src = [ "lib/some_lib/setup.py" ] `which get passed to black's `find_project_root` which will find the `pyproject.toml` file in `lib/some_lib/pyproject.toml` and you will load that, which has no settings for `darker`. It would be good to revert this or at least allow the option to configure the path to the `darker` config file as well.

akaihola:
Darker's own configuration is now read from the `PATH` TOML file when `-c PATH` or `--config PATH` is used.

Fixes #244 (partially).

- [x] accept file for `PATH`
- [x] unit tests for file as `PATH`
- [x] accept directory for `PATH`
- [x] unit tests for directory as `PATH`
- [x] change log
- [x] review

### DIFF
--- a/src/darker/config.py
+++ b/src/darker/config.py
@@ -122,8 +122,15 @@ def load_config(path: Optional[str], srcs: Iterable[str]) -> DarkerConfig:
 
     """
     if path:
-        config_path = Path(path)
-        if not config_path.is_file():
+        for candidate_path in [Path(path), Path(path) / "pyproject.toml"]:
+            config_path = Path(candidate_path)
+            if config_path.is_file():
+                break
+        else:
+            if Path(path).is_dir() or path.endswith(os.sep):
+                raise ConfigurationError(
+                    f"Configuration file {Path(path) / 'pyproject.toml'} not found"
+                )
             raise ConfigurationError(f"Configuration file {path} not found")
     else:
         config_path = find_project_root(tuple(srcs or ["."])) / "pyproject.toml"
--- a/src/darker/config.py
+++ b/src/darker/config.py
@@ -131,8 +131,8 @@ def load_config(path: Optional[str], srcs: Iterable[str]) -> DarkerConfig:
     """
     if path:
         for candidate_path in [Path(path), Path(path) / "pyproject.toml"]:
-            config_path = Path(candidate_path)
-            if config_path.is_file():
+            if candidate_path.is_file():
+                config_path = candidate_path
                 break
         else:
             if Path(path).is_dir() or path.endswith(os.sep):
--- a/src/darker/config.py
+++ b/src/darker/config.py
@@ -5,7 +5,7 @@ import os
 import sys
 from argparse import ArgumentParser, Namespace
 from pathlib import Path
-from typing import Iterable, List, cast
+from typing import Iterable, List, Optional, cast
 
 import toml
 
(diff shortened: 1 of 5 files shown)


akaihola:
@DavidCDreher here's the first part for fixing #244:

>### Version 1.5.0
>
>- read Darker configuration from the file specified using `-c`/`--config` again (like it used to be)
>- assume `<directory>/pyproject.toml` if the value of `-c`/`--config` is a directory

I invited you as a collaborator on this repository – accept if you will, and I'll assign you as a reviewer for this PR.

akaihola:
@roniemartinez I added you as a reviewer just in case you'd be interested to take a look at this last remaining PR for the 1.5.0 release.

Also, I'm curious about your opinion to [this plan](https://github.com/akaihola/darker/issues/244#issuecomment-1087647478) for separating command line options which point to Darker, Black and Isort configuration files. This PR is the first part of that plan.

roniemartinez (approved):
LGTM

src/darker/config.py
```diff
@@ -114,23 +114,41 @@ def override_color_with_environment(pyproject_config: DarkerConfig) -> DarkerCon
...
-        pyproject_toml = toml.load(path)
-        config = cast(
-            DarkerConfig, pyproject_toml.get("tool", {}).get("darker", {}) or {}
-        )
-        replace_log_level_name(config)
-        validate_config_output_mode(config)
-        return config
-    return {}
+    if path:
+        for candidate_path in [Path(path), Path(path) / "pyproject.toml"]:
```
MatthijsBurgh, line 133:
```suggestion
        for candidate_path in [path, os.path.join(path, "pyproject.toml")]:
```
Better to use `os.path.join`. Also You are constructing a new `Path` object on the `candidate_path`, so no need to do it here too.

src/darker/config.py
```diff
@@ -114,23 +114,41 @@ def override_color_with_environment(pyproject_config: DarkerConfig) -> DarkerCon
...
+            config_path = Path(candidate_path)
+            if config_path.is_file():
+                break
+        else:
+            if Path(path).is_dir() or path.endswith(os.sep):
+                raise ConfigurationError(
+                    f"Configuration file {Path(path) / 'pyproject.toml'} not found"
+                )
+            raise ConfigurationError(f"Configuration file {path} not found")
```
MatthijsBurgh, line 142: I think this should be indented an extra time

MatthijsBurgh (changes requested):

src/darker/config.py
MatthijsBurgh: Or is this `else` related to the `for`? Then it is good I think.

src/darker/config.py
akaihola: In this code base, I prefer to always use `pathlib` when possible and to avoid `os.path`.

I removed the extra `Path()` construction. Thanks @MatthijsBurgh!

src/darker/config.py
akaihola: Yes, this is one of the rare cases where I find `for..else` useful. And the `if..raise..raise` is just for creating accurate error messages depending on whether a file or directory path was provided. It's not foolproof for every situation, but good enough I think.

src/darker/config.py
MatthijsBurgh, line 133:
```suggestion
        for candidate_path in [Path(path), Path(path, "pyproject.toml")]:
```
I think is the better way to construct the path object.

MatthijsBurgh (changes requested):
Small change

src/darker/config.py
```diff
@@ -114,23 +114,41 @@ def override_color_with_environment(pyproject_config: DarkerConfig) -> DarkerCon
...
-    return {}
+    if path:
+        for candidate_path in [Path(path), Path(path) / "pyproject.toml"]:
+            if candidate_path.is_file():
+                config_path = candidate_path
+                break
+        else:
+            if Path(path).is_dir() or path.endswith(os.sep):
+                raise ConfigurationError(
+                    f"Configuration file {Path(path) / 'pyproject.toml'} not found"
```
akaihola, line 140: Let's build the path using `Path()` arguments here, too then:
```suggestion
                    f"Configuration file {Path(path, 'pyproject.toml')} not found"
```

PR #373, opened 2022-08-27
Show linter invocation in debug level log

akaihola:
Helps debugging #368.


akaihola:
@deadkex would you have time to review this small one as well?

deadkex (approved):
Looks good

PR #376, opened 2022-08-27
Refactoring: Care only if import sorting was done, not whether isort was run or not.

akaihola:
This refactoring simplification is in preparation for unified exclusions (#377), which is in preparation for `flynt` support (#266 & #308).

Merge this before any of those.

PR #308, opened 2022-02-24
Add conversion to f-strings using `flynt`

akaihola:
Fixes #266.

- [x] PoC implementation
- [x] unit tests for:
	- [x] `--help` content
	- [x] actual f-string conversions
- [x] documentation in README
	- [x] command line option in the `--help` copy
	- [x] description
	- [x] "New in" note
- [x] Merge #376 
- [x] Merge #377
- [x] change log

Issue #266, opened 2022-01-30
No support for using `flynt` to convert old string literal formatting to f-strings


jedie:
https://github.com/ikamensh/flynt is a great tool. Think it's nice if darker includes this.

akaihola:
Great idea! It would be a good addition to Black and isort. Let's see how it fits in – it was a bit tricky to get isort in since it can change the order of lines, but that shouldn't be an issue here. It's more that we should be careful about in which order different tools are applied.

PR #377, opened 2022-08-27
Refactor Black and isort file exclusions into one data structure

akaihola:
This is in preparation for flynt support (#266 & #308).

Based on #376, merge that one before this PR.

### DIFF
--- a/src/darker/__main__.py
+++ b/src/darker/__main__.py
@@ -440,7 +440,9 @@ def main(argv: List[str] = None) -> int:
                 path.relative_to(root) for path in files_to_process
             }
         black_exclude = {
-            str(f) for f in changed_files_to_process if root / f not in files_to_blacken
+            str(path)
+            for path in changed_files_to_process
+            if root / path not in files_to_blacken
         }
 
     use_color = should_use_color(config["color"])
--- a/src/darker/__main__.py
+++ b/src/darker/__main__.py
@@ -19,7 +19,7 @@ from darker.black_diff import (
 from darker.chooser import choose_lines
 from darker.command_line import parse_command_line
 from darker.concurrency import get_executor
-from darker.config import OutputMode, dump_config
+from darker.config import Exclusions, OutputMode, dump_config
 from darker.diff import diff_chunks
 from darker.exceptions import DependencyError, MissingPackageError
 from darker.git import (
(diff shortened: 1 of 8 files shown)


akaihola:
@jedie, here's another refactoring which was done in preparation for flynt support. Could you review this one as well?

src/darker/__main__.py
```diff
@@ -132,17 +131,16 @@ def _isort_and_blacken_single_file(  # pylint: disable=too-many-arguments
...
-    else:
-        rev2_isorted = rev2_content
-    if relative_path_in_rev2 not in black_exclude:
+    rev2_isorted = apply_isort(
+        rev2_content,
+        relative_path_in_rev2,
+        exclude.isort,
+        edited_linenums_differ,
+        black_config.get("config"),
+        black_config.get("line_length"),
```
jedie, line 140: Maybe pass the complete `black_config` object and the function can use the values he needs?

src/darker/__main__.py
```diff
@@ -442,7 +440,7 @@ def main(argv: List[str] = None) -> int:
                 path.relative_to(root) for path in files_to_process
             }
         black_exclude = {
-            f for f in changed_files_to_process if root / f not in files_to_blacken
+            str(f) for f in changed_files_to_process if root / f not in files_to_blacken
```
jedie, line 443: Maybe rename `f` to `path` ?

src/darker/tests/test_import_sorting.py
```diff
@@ -55,7 +55,43 @@ def test_apply_isort(git_repo, encoding, newline, content, expect):
...
+        content_, src, exclude=[], edited_linenums_differ=edited_linenums_differ
+    )
+
+    assert result.lines == expect
+    assert result.encoding == encoding
+    assert result.newline == newline
+
+
+@pytest.mark.parametrize("encoding", ["utf-8", "iso-8859-1"])
+@pytest.mark.parametrize("newline", ["\n", "\r\n"])
```
jedie, line 68: maybe add `\r` from Mac? `["\n", "\r\n", "\r"]`

src/darker/__main__.py
akaihola: I've considered that, but the rest of `black_config` is Black specific configuration, and we're calling the `apply_isort()` function here. I liked to be explicit here.

Maybe an `IsortOptions` dataclass could encapsulate both of these instead.

Anyway, I decided to not touch this right now.

src/darker/__main__.py
akaihola: Thanks, I renamed that.

src/darker/tests/test_import_sorting.py
akaihola: I started adding such test cases (there are lots of these across the test suite), and started getting test failures. See the [mac-newline-tests](akaihola/darker/tree/mac-newline-tests) branch.

However, I haven't yet heard of any complaints from macos users about newlines being treated wrong. Are you sure about `\r`?

PR #391, opened 2022-09-04
Apply more granular changes by using `# fmt: skip`

### LINKED ISSUE #388: Darker changes lines when the previous lines got changed
**Describe the bug**
- Darker changes lines when the previous lines got changed.
- It happens when there are trailing closing parenthesis of the function call.

**To Reproduce**

[darker_tests.zip](https://github.com/akaihola/darker/files/9446046/darker_tests.zip)

```
unzip darker_tests.zip
cd darker_tests
git --no-pager diff
darker -l 79 spam.py --diff
```

**Expected behavior**
- Only the line that was changed by editing gets updated. (in the below screenshot, only the first call of func1)

**Screenshots**
image

**Environment (please complete the following information):**
 - OS: macOS
 - Python version: 3.9.10
 - Git version [e.g. 2.36.0]
```
% git --version
git version 2.32.1 (Apple Git-133)
hub version refs/heads/master
```
 - Darker version [e.g. 1.5.0]
```
% darker --version
1.5.0
```
 - Black version [e.g. 22.3.0]
```
% black --version
black, 22.1.0 (compiled: yes)
```

wkentaro:
Close https://github.com/akaihola/darker/issues/388

@akaihola What do you think of this solution? (NOTE: this won't work with nonstandard indentations)

**Before**

image

**After**

image

**Limitations**

Screen Shot 2022-09-05 at 2 42 32

### DIFF
--- a/src/darker/__main__.py
+++ b/src/darker/__main__.py
@@ -10,6 +10,8 @@ from difflib import unified_diff
 from pathlib import Path
 from typing import Collection, Generator, List, Tuple
 
+from black.comments import FMT_SKIP
+
 from darker.black_diff import (
     BlackConfig,
     filter_python_files,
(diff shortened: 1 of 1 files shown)


akaihola:
@wkentaro indeed an interesting approach!

In some test cases, the end result seems to be different, see [build #1419](https://github.com/akaihola/darker/runs/8178606957?check_suite_focus=true). Do you think these are in fact improvements to old behavior instead of erroneous results due to a bug?

What happens if the decorated lines become longer than 88 characters (or whatever is the configured Black margin)? I suppose that's ok because the comment instructs Black to ignore those lines and pass them through as such.

Handling of empty lines seems to be asymmetrical in your implementation – decorating them is probably supposed to be special cased as well?

What happens inside multi-line strings?

What happens if I modify only some lines in a multi-line expression which doesn't already conform to Black formatting rules? For example if this is the old revision:
```python
call_func(first_arg=1,
    second_arg = 2,
    third_arg="three"
)
```
and this is my new revision:
```python
call_func(first_arg=1,
    second_arg = 2,
    third_arg=3
)
```
(so only the `third_arg=` line changed)

Would the whole function call still be reformatted by Black, or only the line I modified? Which one would be desirable in your opinion?

Could you approach replace the current `difflib`-chunk-applying approach completely?

wkentaro:
> What happens if the decorated lines become longer than 88 characters (or whatever is the configured Black margin)? I suppose that's ok because the comment instructs Black to ignore those lines and pass them through as such.

Yes, that's going to be ok.

---

> Handling of empty lines seems to be asymmetrical in your implementation – decorating them is probably supposed to be special cased as well?

When there is a line `  # fmt: skip` (with 2 spaces), Black reformats it to `# fmt: skip` (no spaces), that's why we need to have an asymmetric de-decoration.

---

> What happens inside multi-line strings?

In this case, Black ignores `# fmt:skip` decoration, and the decoration `# fmt:skip` is treated as a comment so will affect the  result.

image

If we apply black with 2 times, it gives an expected result.

image

---

> Would the whole function call still be reformatted by Black, or only the line I modified? According to your "After" example, the latter seems true. Which one would be desirable in your opinion?

In this particular case, the Black won't reformat anything as there is going to be `)  # fmt: skip`, and it cancels the whole reformat of the function call. This is probably the big drawback of this approach.

In my particular use case (a codebase of several people changing), I prefer the more conservative result since I can still reformat lines if I need it by `black --code "<lines extracted by editor>"` (I use `vim` and its visual selection). Reformatting again is easier than reverting a reformat.

---

> Could you approach replace the current difflib-chunk-applying approach completely?

I just found an edge case by trying this. Without `difflib`, it creates an indented line like below:

image

this is because `# fmt:skip` will be formatted `  # fmt:skip` and the trailing 2 spaces cannot be removed with the current algorithm.

---

> Could we avoid the # fmt:skip variants limitation by appending something unique as another comment? Would Black still recognize its own skip directive?

Yes, it seems possible by changing both `FMT_SKIP` and `FMT_PASS`.

akaihola:
>I prefer the more conservative result [...] Reformatting again is easier than reverting a reformat.

This seems to be a matter of preference. For me it's the opposite – I want the code base to be marching steadily towards complete Blackification, and if this sometimes means that a couple of neighboring unmodified statements are reformatted as well, that's a reasonable trade-off.

This is what I attempted to solve in #221. I hoped to be able to extract Black's reformatted output in chunks corresponding to expressions which are parenthetically balanced and cover a sequence of complete lines of code. If it could also tell which original lines were reformatted, that set of information could replace the use of `difflib` altogether. Unfortunately this approach didn't really work out (see [discussion](/akaihola/darker/issues/388#issuecomment-1230707491)).

akaihola:
I'm postponing to consider this for 1.8.0. There are other features lined up for 1.7.0 and I prefer to not delay it further.

wkentaro:
it's stale, so closing now. (so that it won't show in my github pull request dashboard).

Issue #380, opened 2022-08-28
Point out shortcomings in `--lint mypy`


akaihola:
- [ ] in `README.rst`, recommend against using `--lint mypy`
- [ ] warn about and explain the current shortcoming (`mypy` errors in other lines or files than the modified line are not included)
- [ ] possibly additionally drop other mentions of `mypy`

_Originally posted by @akaihola in https://github.com/akaihola/darker/discussions/328#discussioncomment-2367260_

akaihola:
#393 will fix Mypy behavior soon, so maybe not add this after all?

akaihola:
#393 added baseline linting, so this is not needed.

PR #393, opened 2022-09-12
Run linters twice: 1) baseline, 2) current; then show new messages

### LINKED ISSUE #383: Track distant side effects of changes by doing an extra baseline run for linters
**Problem:** Darker only shows linter erros on modified lines. But e.g. when correcting a function signature, `mypy` easily finds the static typing errors in invocations on unmodified lines elsewhere in the same or other `.py` files.

**Solution:** Check out `rev1`, run the linters, and record their outputs as the _baseline_. Check out `rev2`, run them again, and record the result. Similar to Black and Flynt processing, parse `git diff` output. Hide _only_ linter messages which
1. fall on an `equal` chunk, and
1. also appear in the recorded baseline with an identical messages and a line number corresponding to the chunk's location in `rev1`.

This still won't solve code moves between files, but it's still vastly better than the current algorithm.

**Background:**

We found prior art in [basedmypy](/KotlinIsland/basedmypy) and [flakehell baseline](https://flakehell.readthedocs.io/commands/baseline.html).

It seems that baseline support in `basedmypy` is still rather primitive – it uses a heuristic of scanning for the same type checking error within ±50 lines, which can of course lead to false positives/negatives. We should make use of information from `git diff` gives about moved code, as Darker already does when filtering modifications from Black (and soon Flynt).

I suspect the same applies to `flakehell baseline` since the baseline file is just a list of hashes, and it doesn't use a diff to track code moves.

**Related:** We're documenting the current shortcomings in #380.

_Originally posted by @akaihola in https://github.com/akaihola/darker/discussions/328#discussioncomment-3492254_

akaihola:
Hide linter messages which were already present in the baseline. Track unmodified blocks of code which are moved up or down due to other changes, and match linter messages even though the line number might have changed since the baseline.

Fixes #383 and #380.

- [x] Clean up commits of the branch
- [x] Documentation
- [x] Fix `cov_to_lint.py` support
- [x] Fix Windows Heisenbug: tests fail unless debug logging is enabled (see #442 for comparison)
- [x] Show also old linter errors which fall on modified lines
- [x] Change log
- [x] Merge #444 
- [x] Rebase
- [x] Self-review
- [ ] Peer review

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,6 +12,10 @@ Added
   ``--target-version`` command line option
 - In ``README.rst``, link to GitHub searches which find public repositories that
   use Darker.
+- Linters are now run twice: once for ``rev1`` to get a baseline, and another time for
+  ``rev2`` to get the current situation. Old linter messages which fall on unmodified
+  lines are hidden, so effectively the user gets new linter messages introduced by
+  latest changes, as well as persistent linter messages on modified lines.
 
 Fixed
 -----
--- a/README.rst
+++ b/README.rst
@@ -31,8 +31,13 @@ What?
 =====
 
 This utility reformats and checks Python source code files.
-However, when run in a Git repository, it only applies reformatting and reports errors
-in regions which have changed in the Git working tree since the last commit.
+However, when run in a Git repository, it compares an old revision of the source tree
+to a newer revision (or the working tree). It then
+
+- only applies reformatting in regions which have changed in the Git working tree
+  between the two revisions, and
+- only reports those linting messages which appeared after the modifications to the
+  source code files.
 
 The reformatters supported are:
 
@@ -726,7 +731,8 @@ It defaults to ``"--check --diff --color"``.
 You can e.g. add ``"--isort"`` to sort imports, or ``"--verbose"`` for debug logging.
 
 To run linters through Darker, you can provide a comma separated list of linters using
-the ``lint:`` option. Only ``flake8``, ``pylint`` and ``mypy`` are supported.
+the ``lint:`` option. Only ``flake8``, ``pylint`` and ``mypy`` are supported. Other
+linters may or may not work with Darker, depending on their message output format.
 Versions can be constrained using ``pip`` syntax, e.g. ``"flake8>=3.9.2"``.
 
 *New in version 1.1.0:*
@@ -745,7 +751,9 @@ The ``lint:`` option.
 Using linters
 =============
 
-One way to use Darker is to filter linter output to modified lines only.
+One way to use Darker is to filter linter output to only those linter messages
+which appeared after the modifications to source code files,
+as well as old messages which concern modified lines.
 Darker supports any linter with output in one of the following formats::
 
     <file>:<linenum>: <description>
@@ -892,7 +900,9 @@ are applied to the edited file.
 Also, in case the ``--isort`` option was specified,
 isort_ is run on each edited file before applying Black_.
 Similarly, each linter requested using the `--lint <command>` option is run,
-and only linting errors/warnings on modified lines are displayed.
+and only those linting messages are displayed which appeared after the modifications to
+the source code files,
+or which were there already before but now fall on modified lines.
 
 
 License
--- a/src/darker/__main__.py
+++ b/src/darker/__main__.py
@@ -350,11 +350,14 @@ def main(  # pylint: disable=too-many-locals,too-many-branches,too-many-statemen
     8. verify that the resulting reformatted source code parses to an identical AST as
        the original edited to-file
     9. write the reformatted source back to the original file
-    10. run linter subprocesses for all edited files (10.-13. optional)
-    11. diff the given revision and worktree (after isort and Black reformatting) for
-        each file reported by a linter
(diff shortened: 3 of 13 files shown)


akaihola:
@PatrickJordanCongenica, you may be interested to try out this branch! It fixes #383 (originally raised by you in #328).

Would you be interested to do a code review as well? Would you accept an invite as a collaborator on the repository so you could be assigned as a reviewer?

I'm going to clean up some commits still though before this is ready for review.

akaihola:
Hm, let's rethink a bit. With this patch, Darker is only going to show those linter messages which were newly introduced as a result of changes in the code.

But we'd probably want it to also show old linter messages if they happen to fall on modified lines!

akaihola:
>we'd probably want it to also show old linter messages if they happen to fall on modified lines

I believe the patch now does this as well.

akaihola:
# **HEISENBUG!**

Tests fail on Windows unless debug logging is enabled (see #442).

This test:
```python
    def test_check_linter_output(tmp_path):
        """``_check_linter_output()`` runs linter and returns the stdout stream"""
>       with linting._check_linter_output(
            "echo",
            tmp_path,
            {Path("first.py"), Path("second.py")},
            {},
        ) as stdout:
```

calls `_check_linter_output()`:
```python
@contextmanager
def _check_linter_output(
    cmdline: str,
    root: Path,
    paths: Collection[Path],
    env: Dict[str, str],
) -> Generator[IO[str], None, None]:
    cmdline_and_paths = shlex.split(cmdline) + [str(path) for path in sorted(paths)]
    with Popen(
        cmdline_and_paths,
        stdout=PIPE,
        encoding="utf-8",
        cwd=root,
        env=env,
    ) as linter_process:
```

`cmdline_and_paths` should at this point equal something like
```python
["echo", "C:\\tmp\\path12345"]
```

`Popen.__init__()` in turn calls `Popen`s `def _execute_child(self, args, executable, ...)` but we get this error:
```python
        # Start the process
        try:
>           hp, ht, pid, tid = _winapi.CreateProcess(executable, args, ...)
E                                    OSError: [WinError 87] The parameter is incorrect
```

Looking at it with the debugger, it seems `executable is None`. In #442, I'm trying to find out why, but since the error disappears when turning on debug logging, I can't.

Good ideas appreciated!

akaihola:
The Heisenbug might be about `echo` being an [internal command](https://www.computerhope.com/jargon/i/intecomm.htm), see the question [Cannot find the file specified when using subprocess.call('dir') in Python](https://stackoverflow.com/questions/49709826/cannot-find-the-file-specified-when-using-subprocess-calldir-in-python) on Stack Overflow. But why would it work when debug logging is enabled?!?

akaihola:
I was considering running linters through a shell (`Popen(..., shell=True)`), but that would require quoting paths passed on the linter command line and also be less secure. So instead I'm trying to just run `cmd /c echo` instead of just `echo` in tests if running on Windows.

akaihola:
`cmd /c echo` didn't help. Still `executable is None` even though
```python
self = <Popen: returncode: None args: ['cmd', '/c', 'echo', 'first.py', 'second.py']>
args = 'cmd /c echo first.py second.py', executable = None, preexec_fn = None
```

akaihola:
Probably the "Warning" section in [Popen Constructor](https://docs.python.org/3/library/subprocess.html#popen-constructor) documentation is relevant here. Maybe Darker needs to find the `executable` on Windows?

akaihola:
I think I got it! It's because we have `env=env` in
```python
    with Popen(  # nosec
        cmdline_and_paths,
        stdout=PIPE,
        encoding="utf-8",
        cwd=root,
        env=env,
    ) as linter_process:
```

akaihola:
The other test has this problem:
```python
(Pdb) cmdline
'type C:\\Users\\akaihola\\AppData\\Local\\Temp\\2\\pytest-of-akaihola\\pytest-7\\test_run_linters_line_separati0\\dummy-linter-output.txt'
(Pdb) shlex.split(cmdline)
['type', 'C:UsersakaiholaAppDataLocalTemp2pytest-of-akaiholapytest-7test_run_linters_line_separati0dummy-linter-output.txt']
```

akaihola:
@PatrickJordanCongenica, baseline linting is now ready for review!

akaihola:
The [well-known](https://github.com/psf/requests-html/issues/249#issuecomment-480772748) Python 3.7.2 [bug #35797](https://bugs.python.org/issue35797) doesn't explain the test failure on `windows-latest, 3.7` (see [build #1709](https://github.com/akaihola/darker/actions/runs/3859025867/jobs/6578213853)) – we're running 3.7.9:
```yaml
Run actions/setup-python@v4
  with:
    python-version: 3.7
    check-latest: false
    token: ***
    update-environment: true
Installed versions
  Successfully set up CPython (3.7.9)
```

```python
> os.unlink(fullname)
E PermissionError: [WinError 5] Access is denied:
  'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\tmp12s99vo_\\test_run_linters_messages_befo1\\.git\\objects
   \\7a\\ea9bf550cfe9c14baf1e82fd65ebf936d81e22'

PR #408, opened 2022-11-14
Add test cases for mac-style newlines

akaihola:
See https://github.com/akaihola/darker/pull/377#discussion_r957014910


akaihola:
Postponing to version 1.7.1 since we haven't heard any requests about this.

PR #411, opened 2022-12-05
support target-version

aljazerzen:
Adds support for target-version in pyproject.toml

I'm not sure how to add (or even run) tests, so help on that would be appreciated.

Also, I've ran black and it formats quite a few code chunks. I'm I running it right?

- [x] Update `README`
- [x] Update the change log
- [x] Review

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,6 +4,7 @@ requires = ["setuptools", "wheel"]  # PEP 508 specifications.
 
 [tool.black]
 skip-string-normalization = false
+target-version = ["py311"]
 
 [tool.isort]
 profile = "black"
--- a/src/darker/black_diff.py
+++ b/src/darker/black_diff.py
@@ -72,10 +72,12 @@ DEFAULT_INCLUDE_RE = re_compile_maybe_verbose(DEFAULT_INCLUDES)
 
 class BlackConfig(TypedDict, total=False):
     """Type definition for Black configuration dictionaries"""
+
     config: str
     exclude: Pattern[str]
     extend_exclude: Pattern[str]
     force_exclude: Pattern[str]
+    target_version: str
     line_length: int
     skip_string_normalization: bool
     skip_magic_trailing_comma: bool
@@ -83,6 +85,7 @@ class BlackConfig(TypedDict, total=False):
 
 class BlackModeAttributes(TypedDict, total=False):
     """Type definition for items accepted by ``black.Mode``"""
+
     target_versions: Set[TargetVersion]
     line_length: int
     string_normalization: bool
@@ -114,6 +117,17 @@ def read_black_config(src: Tuple[str, ...], value: Optional[str]) -> BlackConfig
     ]:
         if key in raw_config:
             config[key] = raw_config[key]  # type: ignore
+    if "target_version" in raw_config:
+        target_version = raw_config["target_version"]
+        if isinstance(target_version, str):
+            config["target_version"] = target_version
+        elif isinstance(target_version, list):
+            # Convert TOML list to a Python set
+            config["target_version"] = set(target_version)
+        else:
+            raise ConfigurationError(
+                f"Invalid target-version = {target_version!r} in {value}"
+            )
     for key in ["exclude", "extend_exclude", "force_exclude"]:
         if key in raw_config:
             config[key] = re_compile_maybe_verbose(raw_config[key])  # type: ignore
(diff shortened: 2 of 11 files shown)


akaihola:
Thanks @aljazerzen for the pull request!

For the curious, we're talking about this Black option:
```
  -t, --target-version [py33|py34|py35|py36|py37|py38|py39|py310|py311]
                                  Python versions that should be supported by
                                  Black's output. [default: per-file auto-
                                  detection]
```

@aljazerzen , to run tests,
- clone the Darker repository,
- create an empty virtualenv, 
- do a develop install with testing dependencies (`pip install -e '.[test]'` from repository root), and
- run the tests: `pytest`

Some of the existing tests do seem to require modifications. See te failing checks in this PR.

Don't run Black on the Darker code base. Instead, run `darker --isort --revision master... .` in your branch to just reformat the lines you've modified. Darker is eating its own dogfood by having some non-Black-formatted bits in its source code.

aljazerzen:
FYI: tests were failing due because my version of git does not have this flag:
```
git init --initial-branch=master
```

akaihola:
The Git version problem will be fixed in #393.

aljazerzen:
I've fixed the tests and reformatted. Consider merging.

There is also a gitignore tweak, which I can remove if too obtrusive.

.gitignore
```diff
@@ -4,3 +4,6 @@ darker.egg-info/
 __pycache__/
 .mypy_cache/
 *.pyc
+/.vscode/
+/build
+/venv
```
akaihola, line 9: Let's keep the newline at the end of this file.

src/darker/help.py
```diff
@@ -89,4 +89,6 @@
 
 LINE_LENGTH = "How many characters per line to allow [default: 88]"
 
+TARGET_VERSION = "Python versions that should be supported. [default: py37]"
```
akaihola, line 92: It would be good to mention this only affects Black and not isort, Mypy, Pylint, Flake8 (or other linters), nor Flynt (support for Flynt is upcoming, see #308). In `black --help` I see the following, maybe use the same wording, so just add "by Black's output"?
```
  -t, --target-version [py33|py34|py35|py36|py37|py38|py39|py310|py311]
                                  Python versions that should be supported by
                                  Black's output. [default: per-file auto-
                                  detection]
```

src/darker/black_diff.py
```diff
@@ -174,6 +178,11 @@ def run_black(src_contents: TextDocument, black_config: BlackConfig) -> TextDocu
     mode = BlackModeAttributes()
     if "line_length" in black_config:
         mode["line_length"] = black_config["line_length"]
+    if "target_version" in black_config:
+        ver_str = black_config["target_version"]
+        ver = next((v for v in TargetVersion if v.name.lower() == ver_str), None)
+        if ver:
+            mode["target_versions"] = {ver}
```
akaihola, line 185: This only adds support for `target-version =` in the configuration file, but doesn't add a corresponding `--target-version=` command line argument. This is inconsistent with other supported Black configuration options. See `make_argument_parser()` in `darker.command_line`.

pyproject.toml
```diff
@@ -4,6 +4,7 @@ requires = ["setuptools", "wheel"]  # PEP 508 specifications.
 
 [tool.black]
 skip-string-normalization = false
+target-version = ["py311"]
```
akaihola, line 7: Does this mean that if when apply Black to Darker's own code base, it might become incompatible with Python versions older than 3.11?

aljazerzen: I'm don't think so, because 3.x python versions are backward compatible (as far as I know). What this does do is explicitly state we want support for newer constructs (like `match`).

aljazerzen:
I don't have a clue why some tests are failing, I hope you can look into it.

akaihola:
I've been working on #393 and finally got some Windows compatibility problems solved. Will get back to this PR next.

akaihola:
And hey @aljazerzen if you have lots of extra times in your hands, feel free to pick any ones of #444, #393, #445, #308, #432 and #446 for review :smirk:

aljazerzen:
Actually, I don't... I needed this option for a codebase that sets python 3.11 version in pyproject.toml. I'm glad post my solution as PR, to extend it to command line and fix it up, but I don't have time for other PRs :cry:

akaihola:
Fair enough, thanks for this contribution! I'll take over this PR, rebase it on `master` and fix the test issues. Have an excellent beginning of year 2023!

aljazerzen:
To you too, thank you for darker!

akaihola:
@aljazerzen I added quite a bunch of tests and did a number of tweaks. Also, I think the `--help` text was wrong and the default value for `target-version` is detected from each `.py` file and not fixed to `py311`.

Could you at least test this branch, and perhaps review a bit before I merge?

aljazerzen:
> default value for target-version is detected from each .py file and not fixed to py311

Oh, you are probably right.

---

I've tested the branch, I approve.

akaihola:
I moved the `.gitignore` update to a separate housekeeping branch, and merged the rest. Thanks a lot @aljazerzen for your contribution!

PR #415, opened 2022-12-17
Add support for Python 3.11

hugovk:
Python 3.11 was [released on 2022-10-24](https://discuss.python.org/t/python-3-11-0-final-is-now-available/20291?u=hugovk) 🚀 

[image](https://discuss.python.org/t/python-3-11-0-final-is-now-available/20291?u=hugovk)

Also test Python 3.12-dev.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -89,6 +89,7 @@ jobs:
           - '3.9'
           - '3.10'
           - '3.11'
+          - '3.12-dev'
         constraints: ['black==22.12.0']
         include:
           - os: ubuntu-latest
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -88,13 +88,14 @@ jobs:
           - '3.8'
           - '3.9'
           - '3.10'
+          - '3.11'
         constraints: ['black==22.12.0']
         include:
           - os: ubuntu-latest
             python-version: '3.7'
             constraints: '--constraint constraints-oldest.txt'
           - os: ubuntu-latest
-            python-version: '3.10'
+            python-version: '3.11'
             constraints: '--constraint constraints-future.txt'
             upgrade: '--upgrade --upgrade-strategy=eager'
     needs:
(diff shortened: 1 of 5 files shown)


akaihola:
You shouldn't directly update `CONTRIBUTORS.rst`. Instead, modify `contributors.yaml` and run the script

    python release_tools/update_contributors.py generate --token=ghp_<your GitHub API token> -r -c

I'll need to document this and add a comment in generated content (`CONTRIBUTORS.rst` and inside `README.rst`).

.github/workflows/python-package.yml
```diff
@@ -88,13 +88,15 @@ jobs:
           - '3.8'
           - '3.9'
           - '3.10'
+          - '3.11'
+          - '3.12-dev'
```
akaihola, line 92: I wonder if using a development version of Python will lead to frequent test failures.

CHANGES.rst
```diff
@@ -369,6 +371,7 @@ Added
 - Initial implementation
 
 .. _Unreleased: https://github.com/akaihola/darker/compare/1.5.0...HEAD
```
akaihola, line 373:
```suggestion
.. _Unreleased: https://github.com/akaihola/darker/compare/1.5.1...HEAD
```

akaihola (commented):
Just a couple of minor comments, almost ready to merge.

.github/workflows/python-package.yml
hugovk: Failures are possible but hopefully not frequent. I've rarely seen them when testing dev versions, but I have found a release blocker in CPython itself by doing so :)

This may also help find early issues in Black and isort, and other linters.

It also means Darker can be ready for 3.12 sooner.

If there are failures, and they're not quick to fix, I wouldn't hesitate to comment this out until it's resolved.

I guess as the linters used by Darker do AST things, there's more exposure to syntax changes before the [beta feature freeze in May](https://peps.python.org/pep-0693/#schedule), so one approach could be to wait until then.

akaihola:
Thanks @hugovk!

hugovk:
You're welcome, and thank you for Darker!

Issue #426, opened 2022-12-21
Document how to specify Black version when using a pre-commit hook


akaihola:
>@victorcui96, I think each pre-commit hook installs into its own environment so modifying your project's requirements.txt will have no effect. The above referencing issue has a fix though, so applying that in your pre commit config would look like:
>
>```
>  - repo: https://github.com/akaihola/darker
>    rev: 1.5.1
>    hooks:
>      - id: darker
>        args:
>           - --line-length=120
>        language_version: python3
>        additional_dependencies: [black==22.10.0]
>```

_Originally posted by @leej3 in https://github.com/akaihola/darker/issues/412#issuecomment-1350785028_

Issue #425, opened 2022-12-21
Pin Black version in the Darker GitHub Action


akaihola:
>But i also think: Darker should set the black version, isn't it?

_Originally posted by @jedie in https://github.com/akaihola/darker/issues/412#issuecomment-1348264321_

This would indeed avoid breakage like the one reported in #412. On the other hand, Action users wouldn't benefit from Black updates even if Darker happens to stay compatible.

akaihola:
This might be irrelevant after #430, but let's consider it when the time comes for version 1.8.0.

akaihola:
Not relevant since #430 was completed.

Issue #412, opened 2022-12-12
Latest version of Black (22.12.0) breaks Darker


wpnbos:
22.12.0 (exception occurs):
```
(venv) william@Sonic ml % pip install black black==22.12.0
Requirement already satisfied: black in ./venv/lib/python3.9/site-packages (22.8.0)
Collecting black
  Downloading black-22.12.0-py3-none-any.whl (167 kB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 167.3/167.3 kB 5.2 MB/s eta 0:00:00
Requirement already satisfied: platformdirs>=2 in ./venv/lib/python3.9/site-packages (from black) (2.5.2)
Requirement already satisfied: pathspec>=0.9.0 in ./venv/lib/python3.9/site-packages (from black) (0.10.1)
Requirement already satisfied: tomli>=1.1.0 in ./venv/lib/python3.9/site-packages (from black) (2.0.1)
Requirement already satisfied: mypy-extensions>=0.4.3 in ./venv/lib/python3.9/site-packages (from black) (0.4.3)
Requirement already satisfied: click>=8.0.0 in ./venv/lib/python3.9/site-packages (from black) (8.1.3)
Requirement already satisfied: typing-extensions>=3.10.0.0 in ./venv/lib/python3.9/site-packages (from black) (4.3.0)
Installing collected packages: black
  Attempting uninstall: black
    Found existing installation: black 22.8.0
    Uninstalling black-22.8.0:
      Successfully uninstalled black-22.8.0
Successfully installed black-22.12.0
(venv) william@Sonic ml % darker
Traceback (most recent call last):
  File "/Users/william/Documents/unetiq/order-link/ml/venv/bin/darker", line 8, in <module>
    sys.exit(main_with_error_handling())
  File "/Users/william/Documents/unetiq/order-link/ml/venv/lib/python3.9/site-packages/darker/__main__.py", line 486, in main_with_error_handling
    return main()
  File "/Users/william/Documents/unetiq/order-link/ml/venv/lib/python3.9/site-packages/darker/__main__.py", line 423, in main
    files_to_process = filter_python_files(paths, root, {})
  File "/Users/william/Documents/unetiq/order-link/ml/venv/lib/python3.9/site-packages/darker/black_diff.py", line 142, in filter_python_files
    gen_python_files(
TypeError: gen_python_files() got an unexpected keyword argument 'gitignore'
```

22.8.0 (works fine):
[...]

Issue #403, opened 2022-11-13
Incompatible with `black-22.10.1.dev19+gffaaf48` and later


akaihola:
**Describe the bug**
Unit tests fail if psf/black@ffaaf48 or later is installed.

**To Reproduce**
1. In a fresh virtualenv, `pip install 'darker[test]==1.5.1' git+https://github.com/psf/black.git@ffaaf48`
2. In a Git working tree of a Python repository, run Darker with `darker`
3. See error

**Expected behavior**
Darker should run successfully and reformat modified regions of Python fules.

**Screenshots**
```python
Traceback (most recent call last):
  File "bin/darker", line 8, in <module>
    sys.exit(main_with_error_handling())
  File "darker/__main__.py", line 486, in main_with_error_handling
    return main()
  File "darker/__main__.py", line 423, in main
    files_to_process = filter_python_files(paths, root, {})
  File "darker/black_diff.py", line 142, in filter_python_files
    gen_python_files(
TypeError: gen_python_files() got an unexpected keyword argument 'gitignore'
```

**Environment (please complete the following information):**
 - OS: NixOS 22.05
 - Python version: 3.9.13
 - Git version: 2.36.2
 - Darker version: 1.5.1
 - Black version: commit `ffaaf48`

**Additional context**
psf/black@ffaaf48 replaces the `gitignore: Optional[PathSpec]` argument of `black.files.gen_python_files()` with `gitignore_dict: Optional[Dict[Path, PathSpec]]`.

Before that commit, `gitignore` is used like:
```python
        # First ignore files matching .gitignore, if passed
        if gitignore is not None and gitignore.match_file(normalized_path):
            report.path_ignored(child, "matches the .gitignore file content")
            continue
```
And after the commit, `gitignore_dict` like:
```python
        # First ignore files matching .gitignore, if passed
        if gitignore_dict and path_is_ignored(child, gitignore_dict, report):
            continue

# [...]

        if child.is_dir():
            # If gitignore is None, gitignore usage is disabled, while a Falsey
            # gitignore is when the directory doesn't have a .gitignore file.
            if gitignore_dict is not None:
                new_gitignore_dict = {
                    **gitignore_dict,
                    root / child: get_gitignore(child),
                }
            else:
                new_gitignore_dict = None

# [then passing `new_gitignore_dict` back to the same function
```

soxofaan:
An out-of-the-box pipx install of darker is also broken at the moment.

[...]

PR #430, opened 2022-12-25
Guard against Black compatibliity breakage

### LINKED ISSUE #382: Guard users against upcoming Black incompatibilities
The process for allowing Darker users to auto-update Black but still guard them against Darker/Black incompatibilities is:

1. The `test-future.yml` workflow regularly tests Darker against the current `main` `HEAD` of Black
1. Make a patch version release as soon as that workflow notifies about an upcoming incompatibility in Black. If that e.g. happened today, we'd release Darker 1.4.3 with a upper limit for the version of the Black dependency to the newest version known to work.
1. Then, after we've fixed the incompatibility, make a new patch version with the fix and the upper version limit removed.

This way anyone specifying `darker~=1.4` for their project, Darker would automatically update to 1.4.3 and prevent Black from updating to an incompatible version before we've released a fixed Darker 1.4.4 or 1.5.0.

- [x] Document `darker~=x.y` as the recommended way to depend on Darker
- [x] Update the `test-future.yml` workflow to point to the guidelines in this issue in case of a failure
- [x] Explain this process in the documentation

_Originally posted by @akaihola in https://github.com/akaihola/darker/discussions/334#discussioncomment-2388545_

akaihola:
Fixes #382.

Also related to
- #403 
- #429
- #426 
- #425 
- #412
- #404 
- #394
- #284
- #270
- #269
- #268 
- #267
- #264
- #262 
- #189 
- #186 
- #183
- #135

### DIFF
--- a/.github/workflows/test-future.yml
+++ b/.github/workflows/test-future.yml
@@ -31,3 +31,57 @@ jobs:
       - name: Test with pytest
         run: |
           pytest
+      - name: Note a possible Black incompatibility and required actions
+        if: failure()
+        shell: python
+        run: |
+          import json
+          import os
+          import urllib.request
+          from distutils.version import LooseVersion
+          from importlib.metadata import version
+          from textwrap import dedent
+
+          for linenum, line in enumerate(open("setup.cfg"), 1):
+              constraint = line.strip()
+              if constraint.startswith("black>="):
+                  column = line.index("black>=") + 1
+                  end_column = len(line)
+                  break
+          else:
+              raise RuntimeError("black>= line not found in setup.cfg")
+
+          response = urllib.request.urlopen(
+              'https://pypi.org/pypi/black/json'
+          ).read().decode()
+          latest_version = max(
+              LooseVersion(s)
+              for s in json.loads(response)['releases'].keys()
+          )
+
+          print(
+              dedent(
+                  f"""
+                  ### :x: Future Black incompatibility? :x:
+
+                  You could add a maximum version constraint for Black on
+                  `setup.cfg` line {linenum}, e.g.
+                  `{constraint},<={latest_version}`
+
+                  See [#382](/akaihola/darker/issues/382)
+                  for more information
+                  """
+              ),
+              file=open(os.getenv("GITHUB_STEP_SUMMARY"), "a"),
+          )
+
+          print(
+              "::notice "
+              "file=setup.cfg,"
+              f"line={linenum},"
+              f"col={column},"
+              f"endColumn={end_column},"
+              "title=Future Black incompatibility?::"
+              "You could add a maximum version constraint for Black here, "
+              f"e.g. {constraint},<={latest_version}"
+          )
--- a/README.rst
+++ b/README.rst
@@ -128,13 +128,33 @@ in Black documentation, or the article
 How?
 ====
 
-To install, use::
+To install or upgrade, use::
 
-  pip install darker
+  pip install --upgrade darker~=1.6.0
 
 Or, if you're using Conda_ for package management::
 
-  conda install -c conda-forge darker isort
+  conda install -c conda-forge darker~=1.6.0 isort
+  conda update -c conda-forge darker
+
+..
+
+    **Note:** It is recommended to use the '``~=``' "`compatible release`_" version
+    specifier for Darker to ensure you have the latest version before the next major
+    release that may cause compatibility issues. If an update to Black introduces an
+    incompatibility with Darker, we will release a first point release for Darker that
+    prevents upgrading Black and a second point release that fixes the incompatibility.
+
+    We test Darker daily against the `Black main branch`_ and strive to proactively
+    fix any potential incompatibilities through this process. If a Black release
+    introduces an incompatibility before the second Darker point release that fixes it,
+    the first Darker point release will downgrade Black to the latest compatible
+    version.
+
+    To be completely safe, you can pin both Darker and Black to known good versions,
+    but this may prevent you from receiving improvements in Black. 
+
+    See issue `#382`_ for more information.
 
 The ``darker <myfile.py>`` or ``darker <directory>`` command
 reads the original file(s),
@@ -159,6 +179,9 @@ You can enable additional features with command line options:
 
 .. _Conda: https://conda.io/
 .. _conda-forge: https://conda-forge.org/
+.. _compatible release: https://peps.python.org/pep-0440/#compatible-release
+.. _Black main branch: https://github.com/psf/black/commits/main
+.. _#382: https://github.com/akaihola/darker/issues/382
 
 
 Example
@@ -571,9 +594,10 @@ do the following:
            hooks:
            -   id: darker
 
-4. install the Git hook scripts::
+4. install the Git hook scripts and update to the newest version::
 
        pre-commit install
+       pre-commit autoupdate
 
 .. _pre-commit: https://pre-commit.com/
 .. _pre-commit Installation: https://pre-commit.com/#installation
@@ -632,7 +656,7 @@ Create a file named ``.github/workflows/darker.yml`` inside your repository with
              options: "--check --diff --color"
              revision: "master..."
              src: "./src"
-             version: "1.6.0"
+             version: "~=1.6.0"
              lint: "flake8,pylint==2.13.1"
 
 There needs to be a working Python environment, set up using ``actions/setup-python``
--- a/README.rst
+++ b/README.rst
@@ -140,21 +140,8 @@ Or, if you're using Conda_ for package management::
 ..
 
     **Note:** It is recommended to use the '``~=``' "`compatible release`_" version
-    specifier for Darker to ensure you have the latest version before the next major
-    release that may cause compatibility issues. If an update to Black introduces an
-    incompatibility with Darker, we will release a first point release for Darker that
-    prevents upgrading Black and a second point release that fixes the incompatibility.
-
-    We test Darker daily against the `Black main branch`_ and strive to proactively
-    fix any potential incompatibilities through this process. If a Black release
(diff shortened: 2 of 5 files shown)


akaihola:
Notifying some contributors who have reported or fixed Black incompatibilities in the past. Maybe some of you are willing to review this one?
@soxofaan @victorcui96 @leej3 @jedie @wpnbos @k-dominik @travishathaway @jezdez @kenodegard @zachnorton4C @flying-sheep @jabesq  @yoursvivek  @Mystic-Mirage  @foxwhite25 @martinRenou  @jenshnielsen @mayk0gan  @MatthijsBurgh

.github/workflows/test-future.yml
```diff
@@ -31,3 +31,10 @@ jobs:
       - name: Test with pytest
         run: |
           pytest
+      - name: Note a possible Black incompatibility and required actions
+        if: failure()
+        run: |
+          echo "### :x: Black incompatibility? :x: " >>$GITHUB_STEP_SUMMARY
+          echo "" >>$GITHUB_STEP_SUMMARY
+          echo "See [#382](/akaihola/darker/issues/382)" >>$GITHUB_STEP_SUMMARY
+          echo "for possible required actions" >> $GITHUB_STEP_SUMMARY
```
akaihola, line 40: How to best test this?

README.rst
```diff
@@ -128,13 +128,24 @@ in Black documentation, or the article
...
+To install or upgrade, use::
 
-  pip install darker
+  pip install --upgrade darker~=1.6.0
 
 Or, if you're using Conda_ for package management::
 
-  conda install -c conda-forge darker isort
+  conda install -c conda-forge darker~=1.6.0 isort
+  conda update -c conda-forge darker
```
akaihola, line 138: I did try with Conda to check that these commands work, but perhaps a Conda expert can confirm?

README.rst
```diff
@@ -128,13 +128,24 @@ in Black documentation, or the article
...
+
+..
+
+    **Note:** We recommend to use the ``~=`` "`compatible release`_" version specifier
+    to ensure you have the newest Darker before the next compatibility-breaking major
+    release. If a Black update introduces an incompatibility with Darker, we'll issue
+    one Darker point release which prevents upgrading Black, and later a second one
+    which fixes the incompatibility.
```
akaihola, line 146: Maybe clarity can be improved here?

README.rst
```diff
@@ -571,9 +584,10 @@ do the following:
            hooks:
            -   id: darker
 
-4. install the Git hook scripts::
+4. install the Git hook scripts and update to the newest version::
 
        pre-commit install
+       pre-commit autoupdate
```
akaihola, line 590: Seems that with `pre-commit` we can't use "compatible release" version specifiers, so it will happily upgrade across major version updates as well. Could a `pre-commit` expert confirm?

README.rst
```diff
@@ -632,7 +646,7 @@ Create a file named ``.github/workflows/darker.yml`` inside your repository with
              options: "--check --diff --color"
              revision: "master..."
              src: "./src"
-             version: "1.6.0"
+             version: "~=1.6.0"
```
akaihola, line 649: This will actually make the GitHub Action install the newest Darker release (before next major release). But will it also cause it to automatically update from time to time?

README.rst
cclauss: You can use `--freeze` https://pre-commit.com/#pre-commit-autoupdate

akaihola: Thanks, nice to learn about those options. However, I guess neither one adds substantial value in this case, and it's best to keep the example simple as is, would you agree?

.github/workflows/test-future.yml
akaihola: Probably simply by introducing a deliberate test failure, and removing it before merging.

akaihola: I now added a failing test in `test_black_diff.py`.

MatthijsBurgh: @akaihola Yes, you could do that. Maybe do it on another branch to prevent "force-push" pollution of this PR.

akaihola: Oops, I already did it here. But I'm anyway doing quite a bit of force-pushing when working on branches, so it doesn't make much of a difference.

README.rst
MatthijsBurgh: What happens if the user has updated Black to a incompatible version? Will the new patch update of Darker downgrade Black?

README.rst
akaihola: Assuming the action will auto-update, either of the following are good setups:
- pin both Darker and Black to exact versions, or
- pin neither so compatibility updates to Darker will be received if a Black update breaks compatibility.

The bad choice is to pin Darker but let Black update, for obvious reasons.

Should these be clarified in the README?

README.rst
akaihola: Yes, that's what should happen. I should probably clarify that.

akaihola: Although as long as the incompatibliity hasn't made in into a Black release, the patch update of Darker will simply prevent Black from upgrading into that release in the future.

README.rst
MatthijsBurgh: By default there is no darker installed in a GH action environment, correct? In that case pip will install the newest available version (matching the version requirements).
Though people can run actions with a docker image. In such case it would be possible darker is already available. Then the `main.py` will not update Darker, when the current version meets the provided version requirements.

README.rst
akaihola: I'd hope that by the time Black releases the incompatibility, we'll already have the second Darker patch update published and be compatible.

.github/workflows/test-future.yml
akaihola: There is now a failing test build [here](https://github.com/akaihola/darker/actions/runs/3782380630/jobs/6430076182), but I can't find the message anywhere.

MatthijsBurgh: It is the `build and test` workflow, not the `test future` workflow. That seems to succeed.

akaihola: ...ah, of course, since `test-future` is only run nightly and not within PRs. We'll have to wait.

```yaml
name: test-future
on:
  schedule:
    - cron: "05 20 * * 6"
```

Also, I guess it would be possible to show a message embedded in the "Files changed" view on the line where we would then need to insert the maximum version constraint for Black.

akaihola: Perhaps like this:
```bash
LINENUM=`sed -n '/black>=/=' setup.cfg|head -1
echo "::notice file=setup.cfg,line=${LINENUM},title=Add Black constraint::You could add a maximum version constraint for Black here"
```

With more magic, we could even find out the latest Black release version number, show `,<=X.Y.Z` and point to the correct column to insert that into...

I guess a code suggestion (as available in review comments) isn't possible, we would need a bot for that.

akaihola: I added that and temporarily enabled `test-future` in the PR so we can see the result sooner.

README.rst
akaihola: But it won't update Black either, so we should be safe except in the unfortunate case where an incompatible Black version was released before we were able to react at all.

README.rst
MatthijsBurgh: Yes, extending the README would indeed be an improvement.

.github/workflows/test-future.yml
akaihola: The notice message worked correctly as can be seen in "Files changed" currently:
kuva

README.rst
MatthijsBurgh: But the current implementation should downgrade Black, in case Black is incompatible with current Darker version

.github/workflows/test-future.yml
akaihola: >With more magic, we could even find out the latest Black release version number, show ,<=X.Y.Z and point to the correct column to insert that into...

I added this as well.

README.rst
akaihola: Yes, I believe it will, if we set e.g. `black>=21.5b1,<=22.12.0` as the requirement in the first patch release before we release a fix.

README.rst
akaihola: README has now been improved.

.github/workflows/test-future.yml
akaihola: And the job summary is visible on the [summary page of the `test-future` workflow run](https://github.com/akaihola/darker/actions/runs/3782833008):
kuva

akaihola: The intentional test failure and the temporary modifications to `test-future.yml` have now been reverted. I'll rebase later to hide them from history as well.

akaihola: Rebased.

MatthijsBurgh (approved):

akaihola:
Thanks for great collaboration @MatthijsBurgh! Have a nice New Year!

MatthijsBurgh:
@akaihola you are welcome. Happy new year to you too.🎆🎇

action/main.py
```diff
@@ -24,6 +24,8 @@
 if VERSION:
     if VERSION.startswith("@"):
         req[0] = f"git+https://github.com/akaihola/darker{VERSION}#egg={req[0]}"
+    elif VERSION.startswith(("~", "=", "<", ">")):
+        req[1] += VERSION
```
foxwhite25, line 28: Shouldn't this be `req[0] += VERSION`? This will cause a IndexError if I am understanding the code correctly.

MatthijsBurgh: I think you are correct. @akaihola, we have missed this one.

akaihola: I remember abandoning the idea to attempt creating unit tests for the action implementation. Would have become handy after all... :smirk:

PR #433, opened 2022-12-26
Instruct how to update contributors

### LINKED ISSUE #416: Document how to update contributors in documentation
You shouldn't directly update `CONTRIBUTORS.rst`. Instead, modify `contributors.yaml` and run the script

    python release_tools/update_contributors.py generate --token=ghp_<your GitHub API token> -r -c

I'll need to document this and add a comment in generated content (`CONTRIBUTORS.rst` and inside `README.rst`).

_Originally posted by @akaihola in https://github.com/akaihola/darker/issues/415#issuecomment-1357794469_

akaihola:
Fixes #416.


akaihola:
@hugovk could you perhaps review this?

CONTRIBUTING.rst
```diff
@@ -34,7 +34,15 @@ To speed up review and increase odds for the PR to be accepted, please
 - verify that mypy static type checking passes
 - document new features or changed behavior in ``README.rst``
 - summarize end-user affecting changes in ``CHANGES.rst``
-- add your information in ``CONTRIBUTORS.rst``
+- add your information in ``contributors.yaml``
+- run
+  ```bash
+  python release_tools/update_contributors.py \
+    generate \
+    --token=ghp_<your GitHub API token> \
```
hugovk, line 42: Can the queries be run without a token, even if it's slower to do so? I expect getting a token is the main blocker for contributors to run this command, and as mentioned, it would put me off contributing. It would be better to do this as part of the release step, rather than PR step, or as automation.

Anyway, if a token is a hard requirement, please link to the place where we need to go to create a token, and list the minimum permissions required.

CONTRIBUTING.rst
```diff
@@ -34,7 +34,15 @@ To speed up review and increase odds for the PR to be accepted, please
...
+    --modify-readme \
+    --modify-contributors
+  ```
 
 GitHub is configured to use Travis CI on each pull request to
```
hugovk, line 47: (Travis CI was replaced by GitHub Actions in https://github.com/akaihola/darker/pull/110.)

CONTRIBUTING.rst
akaihola: When run for the first time with no cached API responses, the script makes one request for each contributor, which may exceed the tokenless API usage limits and throw an exception.

I modified the text to mention that `CONTRIBUTORS.rst` and `README.rst` will be updated at the next release, but that the user may use their own token to update them as well.

I added a section explaining how to create a token. No extra permissions need to be enabled for a fine-grained personal access token.

CONTRIBUTING.rst
akaihola: Fixed that, and mentioned linting, too.

hugovk (approved):
Looks good, thank you!

PR #435, opened 2022-12-26
Remove revision from GitHub Action example

### LINKED ISSUE #410: "Not a valid object name" on GHA
**Describe the bug**
When running on GHA, even with `fetch-depth: 0`, specifying a revision for the "from" point in the range always gives "Not a valid object name <>", e.g. `revision: "master..."` gives "Not a valid object name master"

See [this job](https://github.com/dateutil/dateutil/actions/runs/3609956803/jobs/6083460459) as an example.

**To Reproduce**
1. Create any repository with a master branch
2. Create a github action like so:
```
  darker:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v3
        with:
          fetch-depth: 0
      - uses: actions/setup-python@v4
      - uses: akaihola/darker@1.5.1
        with:
          options: "--check --diff --color --isort"
          revision: "master..."
          src: "./src"
          version: "1.5.1"
```
3. Create a PR.

You can also reproduce this by creating a job that runs `pre-commit` and setting `PRE_COMMIT_FROM_REF` to something like `master`. I have also done:

```
export PRE_COMMIT_FROM_REF=$(git show-ref master)
tox
```

In a job where `TOXENV` is set to an environment that runs `precommit` (and has `passenv=PRE_COMMIT*` set). Same thing happens.
 
**Expected behavior**
I would expect it to run the revision against the base reference.

As it turns out, you can simply *not* specify `revision` to get what I was after in the built-in GHA support (since I was trying to set it to `"${{ github.base_ref }}..."`, but I think it's still a problem in general (plus it means I need to use the built-in GHA support instead of using it integrated with my pre-commit check).

akaihola:
It obtains the commit range automatically.

Fixes #410.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,8 @@ Fixed
 - Fix compatibility with ``black-22.10.1.dev19+gffaaf48`` and later – an argument was
   replaced in ``black.files.gen_python_files()``.
 - Fix tests to work with Git older than version 2.28.x.
+- GitHub Action example now omits ``revision:`` since the commit range is obtained
+  automatically.
 
 
 1.6.0_ - 2022-12-19
--- a/README.rst
+++ b/README.rst
@@ -637,8 +637,7 @@ Create a file named ``.github/workflows/darker.yml`` inside your repository with
          - uses: actions/setup-python@v4
          - uses: akaihola/darker@1.6.0
            with:
-             options: "--check --diff --color"
-             revision: "master..."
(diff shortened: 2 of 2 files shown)


akaihola:
@pganssle, @MatthijsBurgh & @yoursvivek, could one of you perhaps review this documentation change?

MatthijsBurgh (approved):

akaihola:
Thanks once again @MatthijsBurgh.

PR #444, opened 2023-01-04
Fix Git error output

akaihola:
In `_git_check_output()`, stderr output needs to be adapted to work both with `encoding=None` and `encoding="utf-8"`.

Required for #393.

- [x] Update change log
- [x] Review


akaihola:
After sleeping on it, I reviewed the change decided to merge.

PR #462, opened 2023-01-30
Use `lru_cache` clear fixture at test module level

akaihola:


akaihola:
Obsolete, fixed earlier in another PR.

PR #481, opened 2023-03-26
Make string match error verbose in `bump_version.py`

akaihola:
This helps troubleshooting if `CHANGES.rst` patching fails.


akaihola:
`bump_version.py` has been moved to `darkgray-dev-tools`.

PR #486, opened 2023-03-27
Fix local action outside of this repo

### LINKED ISSUE #489: [Darker Action] 1.7.1 is broken due to actions/commit-range
**Describe the bug**

```
> Run akaihola/darker@1.7.1
Error: Can't find 'action.yml', 'action.yaml' or 'Dockerfile' under '/home/runner/work/chemondis/chemondis/.github/actions/commit-range'. Did you forget to run actions/checkout before running your local action?
```

It looks like this is related to https://github.com/akaihola/darker/commit/639ccb7e54ef44d0a0e272c8602f7176c0570cbc

MatthijsBurgh:
Fixes #489 and #485.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -37,7 +37,7 @@ runs:
   steps:
     - name: Commit Range
       id: commit-range
-      uses: akaihola/darker/.github/actions/commit-range@1.7.0
+      uses: akaihola/darker/.github/actions/commit-range@1.7.1
     - name: Run Darker
       run: |
         # Exists since using github.action_path + path to main script doesn't
--- a/action.yml
+++ b/action.yml
@@ -37,7 +37,7 @@ runs:
   steps:
     - name: Commit Range
       id: commit-range
-      uses: ./.github/actions/commit-range
+      uses: akaihola/darker/.github/actions/commit-range@1.7.0
     - name: Run Darker
       run: |
         # Exists since using github.action_path + path to main script doesn't
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,8 @@ Added
 
 Fixed
 -----
+- Revert running ``commit-range`` from the repository itself. This broke the GitHub
+  action.
 
 
 1.7.1_ - 2023-03-26


akaihola:
Hi @MatthijsBurgh, thanks for the patch, and sorry for missing the side effect of my "improvement".

The problem I was solving was to make sure latest changes to the `commit-range` action are always used. If we specify the version, we can only use `commit-range` from past releases of Darker. I wonder what would be a proper way to use `commit-range` from the checkout itself, but only for Darker's own builds.

Could you rebase this PR now that I merged your milestone name fix?

MatthijsBurgh:
@akaihola Yes, I fugured out you tried to achieve that. I have to look into it again to see whether there is another solution to it.

akaihola (approved):

MatthijsBurgh:
@akaihola I just have been busy finishing my thesis and now I am on RoboCup event in Portugal.

roniemartinez (approved):

action.yml
```diff
@@ -37,7 +37,7 @@ runs:
   steps:
     - name: Commit Range
       id: commit-range
-      uses: ./.github/actions/commit-range
+      uses: akaihola/darker/.github/actions/commit-range@1.7.1
```
roniemartinez, line 40: [Question] Is `@1.7.1` here necessary?

akaihola: Good point. Would omitting it use the action from `@main`? I'll try to find out.

akaihola: From [`jobs.<job_id>.uses` documentation]](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_iduses):

>- `{owner}/{repo}/.github/workflows/{filename}@{ref}` for reusable workflows in public and private repositories.
>
>[...]
>`{ref}` can be a SHA, a release tag, or a branch name. Using the commit SHA is the safest for stability and security. For more information, see "[Security hardening for GitHub Actions](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#reusing-third-party-workflows)."

All examples on that documentation page do contain a `@{ref}`.

Using `@main` would cause old versions of the action to call `commit-range` from current `@main`, so effectively the action version specifier wouldn't guarantee that all the code which is run will always stay the same.

Thus, I think using a version number here is the proper thing to do. Or might I be missing something?

roniemartinez: @akaihola `1.7.1` was the version that has the issue if I am not mistaken. So it might be better if we use the proper version after the release.

akaihola: @roniemartinez actually the `commit-range` action wasn't modified in Darker 1.7.1. The only problem is pointing to the action using a same-repository path instead of a full path. This works only in Darker's own workflows, but not when using the Darker action in other repositories.

Fixing this in the `master` branch after each release doesn't feel right.

We could of course keep pointing to a past version (`@1.6.0`, `@1.7.0` and `@1.7.1` are identical) and deal with this dilemma only when there's a need to actually modify `.github/actions/commit-range/`.

PR #487, opened 2023-03-27
(bump_version) Milestones can have an additional comment

MatthijsBurgh:


MatthijsBurgh:
I think the remaining issue is there are multiple variants of a dash. Therefore the first split doesn't do anything.

MatthijsBurgh:
I changed the milestones to use the short dash. But a rerun results in the same behaviour.

MatthijsBurgh:
I changed the milestones to include a short dash. Now the the script does it thing. I can't say it is the correct behaviour, but that is up to you.

MatthijsBurgh:
I changed the milestones to include a short dash. Now the the script does it thing. I can't say it is the correct behaviour, but that is up to you.

akaihola (approved):

akaihola:
Thanks @MatthijsBurgh!

PR #499, opened 2023-05-30
Avoid Mypy error when there's an `__init__.py` in the repo root

### LINKED ISSUE #498: mypy fails with "baseline-revision is not a valid Python package name" in the baseline worktree if the git repo toplevel contains __init__.py
**Describe the bug**

If the top-level of a git repo has an `__init__.py` in it, then the step that does
```shell
git worktree add --quiet --force --force /tmp/tmpF00/baseline-revision HEAD
```
will create a file `baseline-revision/__init__.py`. When mypy sees this, it expects "baseline-revision" to be a valid python package name. But it's not because of the hyphen.

https://github.com/python/mypy/issues/8400 talks a bit about this issue, and there are folks asking for a way to suppress it.

But there's a simple fix in Darker, like

```diff
diff --git a/src/darker/linting.py b/src/darker/linting.py
index bc2055f..1e285d6 100644
--- a/src/darker/linting.py
+++ b/src/darker/linting.py
@@ -524,7 +524,7 @@ def _get_messages_from_linters_for_baseline(

     """
     with TemporaryDirectory() as tmpdir:
-        tmp_path = Path(tmpdir) / "baseline-revision"
+        tmp_path = Path(tmpdir) / "baseline-revision" / root.name
         with git_clone_local(root, revision, tmp_path) as clone_root:
             rev1_commit = git_rev_parse(revision, root)
             result = _get_messages_from_linters(
```

i.e., always create the worktree in a subfolder of "baseline-revision". Sharing the same name as where the repo is checked out should be most reliable. (E.g. if the project config assumes the top-level of the git repo is in PYTHONPATH).

**To Reproduce**
1. Clone this repository: 'https://chromium.googlesource.com/chromiumos/chromite'
    - (or just make a git repo with an `__init__.py` at the top level and any other)
2. Run Darker with `darker -vvv --lint --check --lint "mypy --follow-imports=silent" api/controller/sysroot.py`
3. See error / wrong behavior

"-vvv" will show the output, "baseline-revision is not a valid Python package name".

Otherwise, the effect is just to treat the baseline as having no errors, so all mypy errors on the files are emitted, without filtering.

**Expected behavior**

Just new mypy errors to be emitted.

**Tested with:**

  - darker@a7f586428f44d07795dba601fb21db0fb5ede22b (May 1 2023)
  - mypy 1.3.0
  - Python 3.10.9

akaihola:
Fixes #498.

### DIFF
--- a/src/darker/tests/test_linting.py
+++ b/src/darker/tests/test_linting.py
@@ -4,8 +4,10 @@
 
 import os
 from pathlib import Path
+from subprocess import PIPE, Popen  # nosec
 from textwrap import dedent
-from typing import Dict, Iterable, List, Tuple, Union
+from typing import Any, Dict, Iterable, List, Tuple, Union
+from unittest.mock import patch
 
 import pytest
 
@@ -567,3 +569,35 @@ def test_get_messages_from_linters_for_baseline(git_repo):
         MessageLocation(a_py, 3): [LinterMessage("python", "Third line")],
     }
     assert result == expect
+
+
+class AssertEmptyStderrPopen(
+    Popen  # type: ignore[type-arg]
+):  # pylint: disable=too-few-public-methods
+    # When support for Python 3.8 is dropped, inherit `Popen[str]` instead and remove
+    # the `type-arg` ignore above.
+    """A Popen to use for the following test; asserts that its stderr is empty"""
+
+    def __init__(self, args: List[str], **kwargs: Any):  # type: ignore[misc]
+        super().__init__(args, stderr=PIPE, **kwargs)
+        assert self.stderr is not None
+        assert self.stderr.read() == ""
+
+
+def test_get_messages_from_linters_for_baseline_no_mypy_errors(git_repo):
+    """Ensure Mypy does not fail early when ``__init__.py`` is at the repository root
+
+    Regression test for #498
+
+    """
+    git_repo.add({"__init__.py": ""}, commit="Initial commit")
+    initial = git_repo.get_hash()
+    with patch.object(linting, "Popen", AssertEmptyStderrPopen):
+        # end of test setup
+
+        _ = linting._get_messages_from_linters_for_baseline(
+            linter_cmdlines=["mypy"],
+            root=git_repo.root,
+            paths=[Path("__init__.py")],
+            revision=initial,
+        )
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,9 @@ Fixed
 - Revert running ``commit-range`` from the repository itself. This broke the GitHub
   action.
 - Python 3.12 compatibility in multi-line string scanning.
+- Use the original repository working directory name as the name of the temporary
+  directory for getting the linter baseline. This avoids issues with Mypy when there's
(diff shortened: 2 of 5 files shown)


src/darker/tests/test_linting.py
```diff
@@ -567,3 +580,35 @@ def test_get_messages_from_linters_for_baseline(git_repo):
...
+        self,
+        args: List[str],
+        stdout: int,
+        encoding: str,
+        cwd: Path,
+        env: Dict[str, str],
+    ):
+        super().__init__(
+            args, stdout=stdout, stderr=PIPE, encoding=encoding, cwd=cwd, env=env
+        )
```
tapted, line 597: Not sure if it fits well with your prevailing style (or typechecking - I'm still learning how to bend mypy to my will :p), but I think you could do

```suggestion
    def __init__(self, args: List[str], **kwargs):
        kwargs['stderr'] = PIPE
        super().__init__(args, **kwargs)
```

(this works for Popen because it explicitly declares its trailing parameters to be keyword-only)

src/darker/tests/test_linting.py
```diff
@@ -567,3 +580,35 @@ def test_get_messages_from_linters_for_baseline(git_repo):
...
+        assert self.stderr is not None
+        assert self.stderr.read() == ""
+
+
+def test_get_messages_from_linters_for_baseline_no_mypy_errors(git_repo, capsys):
+    """Test for `linting._get_messages_from_linters_for_baseline`"""
```
tapted, line 603: nit: Should it be something like

```suggestion
    """Ensure mypy does not fail early when __init__.py is at the repo root"""
```

in our codebase we sometimes also mention the bug ID (#498) in a comment somewhere. E.g., 

```
# Regression test for #498.
```

flakes/py312/flake.nix
tapted, line 1: not sure I understand what this file is for (could it be a stray untracked file that got added?)

tapted (approved):
Looks great! Thanks for picking this up so quickly. And sorry for the lag - #timezones, plus I had to learn how to use github codereview haha.

src/darker/tests/test_linting.py
akaihola: I originally had `*args, **kwargs` but Mypy wasn't happy for it. The options were to omit types and do a `#type: ignore`, or to separate and annotate the keyword arguments.

I guess for unit tests and such a minimal helper skipping type checking would be just fine. Would you prefer that?

flakes/py312/flake.nix
akaihola: Removed, accidental. Was related to getting Python 3.12.dev running on my NixOS laptop.

src/darker/tests/test_linting.py
akaihola: Actually this was trivial to do. The class is now much more concise.

akaihola:
Phew, I think I've now dealt with Python<3.9, Mypy, Pylint and Bandit and this can be merged.

I rebased the modifications into the original commits.

Gotta say, despite the trouble, I still do value highly linters, good test coverage and especially code reviews. Thanks @tapted!

PR #509, opened 2023-07-12
Don't check import order using Pylint

akaihola:


akaihola:
This is obsolete, already fixed in another PR earlier.

PR #480, opened 2023-03-26
Fix compatibility with Flynt 0.78 and 1.0.0

akaihola:
Fixes #472 and #541.

We temporarily limited Flynt to version 0.77 and earlier in version 1.7.1 (see #474). This patch fixes compatibility with Flynt 0.78 and 1.0.x.

- [x] merge #501 (to minimize changes on parts which will be changed in that PR)
- [x] rebase on `master`
- [x] drop support for Python 3.7 (strange `flynt` related test failure only on 3.7)

PR #512, opened 2023-07-19
Fix compatibility with Flynt v1.0.0

baod-rate:
Updates the existing flynt-0.78-compatibility branch (https://github.com/akaihola/darker/pull/480) to support, additionally, flynt-1.0.0. It also fixes various issues with the existing flynt 0.78 compatibility changes (e.g. gracefully handling non-installed flynt).

All tests pass locally on Python 3.11, with flynt versions 0.77, 0.78, and 1.0.0, and without flynt installed.

Code formatted with black and isort.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -47,7 +47,7 @@ pygments.lexers =
 
 [options.extras_require]
 flynt =
-    flynt>=0.76,<0.78
+    flynt>=0.76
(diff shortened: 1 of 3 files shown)


akaihola (approved):
Great @baod-rate, I'll include these fixes in the original PR.

PR #515, opened 2023-08-09
Bugfix black .gitignore usage

jedie:
Black didn't use the `.gitignore` file, because `gen_python_files()` will be called with `gitignore_dict=None` and `None` will disable the `.gitignore` usage.

see: https://github.com/akaihola/darker/discussions/323#discussioncomment-6676869

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -61,6 +61,7 @@ test =
     flynt>=0.76,<0.78
     isort>=5.0.1
     mypy>=0.990
+    pathspec  # to test `gen_python_files` in `test_black_diff.py`
     pip-requirements-parser
     pygments
     pytest>=6.2.0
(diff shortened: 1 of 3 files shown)


MatthijsBurgh (approved):
I think this change should indeed change the behaviour, see https://github.com/psf/black/blob/35ce37ded7bd8fdd3950af19e7c11f311ee7b8d8/src/black/files.py#L361-L370

akaihola:
This is great, thanks @jedie and @MatthijsBurgh.

Two questions come to mind:
- Might this change break something if using an old version of Black? If so, the minimum required Black version might need to be bumped. Which Black version does [psf/black@35ce47e](https://github.com/psf/black/blob/35ce37ded7bd8fdd3950af19e7c11f311ee7b8d8/src/black/files.py#L361-L370) affect?
- How to write a unit or integration test for this?

akaihola:
I wrote three unit tests, matching the three different versions of `gen_python_files` in supported Black versions.

Actually `{}` is a valid value only for the most modern versions which use the `gitignore_dict` argument. The older versions with a `gitignore` argument must still get `None` as the value instead. I fixed that too.

akaihola:
It's ok for the future test to fail. Updating Black to 24.2 and beyond is now prevented for Darker 1.7.x, and version 1.8.0 will fix this compatibility issue and allow newer Black versions again.

akaihola (approved):

Issue #490, opened 2023-04-28
Git subprocess call error when too many files in directory, even if gitignore'd


gesslerpd:
**Describe the bug**
On windows, when a repo has too many files in it (e.g. many virtual environments or installed dependencies installed) running `darker` results in the following error:

```
$ python -m darker -i . --rev origin/main --check
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "C:\Users\admin\Documents\flatbracket\venv\Lib\site-packages\darker\__main__.py", line 639, in <module>
    RETVAL = main_with_error_handling()
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\admin\Documents\flatbracket\venv\Lib\site-packages\darker\__main__.py", line 631, in main_with_er
    return main()
           ^^^^^^
  File "C:\Users\admin\Documents\flatbracket\venv\Lib\site-packages\darker\__main__.py", line 575, in main
    changed_files_to_reformat = git_get_modified_python_files(
                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\admin\Documents\flatbracket\venv\Lib\site-packages\darker\git.py", line 474, in git_get_modified_
    changed_paths = _git_diff_name_only(revrange.rev1, revrange.rev2, paths, cwd)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\admin\Documents\flatbracket\venv\Lib\site-packages\darker\git.py", line 434, in _git_diff_name_on
    lines = _git_check_output_lines(diff_cmd, cwd)
  File "C:\Users\admin\Documents\flatbracket\venv\Lib\site-packages\darker\git.py", line 343, in _git_check_output
    return check_output(  # nosec
           ^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\admin\AppData\Local\Programs\Python\Python311\Lib\subprocess.py", line 466, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #542, opened 2024-02-28
Avoid too long Git command lines on Windows [WinError 206]

Svenito:
As per #490 when running darker on Windows with a large codebase, subprocess fails to run with a large list of files.

This occurs because `git_get_modified_python_files` is called with a list of all files that are returned by `filter_python_files`. For a large project this is a long list of files. 

What I attempted to do in this commit is to only process changed files, seeing as this is all we care about anyway. The filtering of the files is moved earlier in the process and is passed the directory or paths set on the commandline, instead of the list of files returned from the `filter_python_files` call. The resulting set is then intersected with all the files Black would process to generate a list of files that should be processed.

While this alleviates the symptom, it could potentially still break if a large amount of files are modified between the current and previous revisions. That said, this is still an improvement on passing the entire list of files to `git_get_modified_python_files`.

### DIFF
--- a/src/darker/tests/test_main.py
+++ b/src/darker/tests/test_main.py
@@ -942,10 +942,7 @@ def test_stdout_path_resolution(git_repo, capsys):
     assert capsys.readouterr().out == 'print("foo")\n'
 
 
-@pytest.mark.skipif(
-    not sys.platform.startswith("win"), reason="Argument length limit on Windows"
-)
-def test_large_file_count(git_repo):
+def test_long_command_length(git_repo):
     # For PR #542 - large character count for changed files
     # on windows breaks subprocess
     # Need to exceed 32762 characters


Svenito:
Note: this isn't passing tests yet. I realise that my changes are munging the file paths and changing behaviour. Working on a better change

akaihola:
Also, as mentioned in #490, are you sure #515 doesn't already fix the problem?

Svenito:
I will test this properly tomorrow but looking at the diff it’s unlikely. I don’t have a problem with gitignore. The issue is that the list of files passed to the git diff call is too large and causes the command to exceed the max length for windows. All those files are Python files so are expected to be there. 
This is a large Django project with a lot of files in sub directory. 

Strange thing I did notice is that installing the wheel from PyPi shows the issue but the pip installed version doesn’t have any issues.

Svenito:
Just tried with the latest master and the error persists

src/darker/git.py
```diff
@@ -110,6 +110,16 @@ def git_is_repository(path: Path) -> bool:
         return False
 
 
+def git_get_repo_root(cwd: Path) -> Path:
+    """Return the root directory of the Git repository containing the given path
+
+    :param cwd: The path to start the search from
+    :return: The root directory of the Git repository, or ``None`` if not found
```
akaihola, line 117: >```
>``None`` if not found
>```
doesn't seem to match what the code does.

It would be good to have unit tests for this code, including a scenario in which no root of a Git repository is found.

src/darker/__main__.py
```diff
@@ -572,9 +573,12 @@ def main(  # pylint: disable=too-many-locals,too-many-branches,too-many-statemen
     else:
         # In other modes, only reformat files which have been modified.
         if git_is_repository(root):
-            changed_files_to_reformat = git_get_modified_python_files(
-                files_to_process, revrange, root
-            )
+            # Get the modified files only.
+            repo_root = git_get_repo_root(root)
+            modified_files = git_get_modified_python_files(paths, revrange, repo_root)
```
akaihola, line 578:
```suggestion
            modified_files = git_get_modified_python_files(files_to_process, revrange, repo_root)
```
Shouldn't this use `files_to_process` to make sure files excluded by Black by default don't get processed?

Svenito: The problem with that is it would cause the same issue again if the command being too long. 
Taking your correction, what it should do is get all changed files and then union with `files_to_process` to make sure only files that are in the latter are kept.

src/darker/__main__.py
```diff
@@ -33,6 +33,7 @@
     get_path_in_repo,
     git_get_content_at_revision,
     git_get_modified_python_files,
+    git_get_repo_root,
```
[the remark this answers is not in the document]
Svenito: Good to know. Thank you

src/darker/git.py
Svenito: Yep. I will add unit tests for this. And also one that reproduces the issue this pr hopes to fix.

src/darker/__main__.py
akaihola: > get all changed files and then union with `files_to_process`

You probably mean intersect instead of union. But yes, this would probably be a good solution.

Can you think of a way to create a test case for this? Generate a large enough number of files in a Git repository and run `main()` on it?

You'll find some tests in `test_command_line.py` and `test_main.py` with examples of helpers to make things happen for such cases.

Svenito: Yeah, sorry, intersect :)

I was going to create a git repo with enough files with a given filename length that exceeds the OS maximum. It's a bit slow, but I can't think of a better way atm to reliably test this

src/darker/git.py
Svenito: I've removed this code in favour of the black compat call.

src/darker/__main__.py
akaihola: The number of files could be reduced by nesting them in a deep file hierarchy with long directory names.

src/darker/tests/test_main.py
```diff
@@ -929,3 +937,22 @@ def test_stdout_path_resolution(git_repo, capsys):
...
+    # on windows breaks subprocess
+    # Need to exceed 32762 characters
+    files = {}
+
+    for _d in range(0, 210):
+        path = "src"
+        for _f in range(0, 4):
+            # Of course windows limits the path length too
+            path = f"{path}/{randomword(30)}"
+        files[f"{path}/{randomword(30)}.py"] = randomword(10)
```
akaihola, line 954: If you first create just one deep directory structure and then put all the 210 files inside the same directory, you'll only need $4 + 210 = 214$ file operations instead of $210 * 4 = 840$ file operations.

src/darker/__main__.py
```diff
@@ -572,9 +573,14 @@ def main(  # pylint: disable=too-many-locals,too-many-branches,too-many-statemen
...
         if git_is_repository(root):
-            changed_files_to_reformat = git_get_modified_python_files(
-                files_to_process, revrange, root
-            )
+            # Get the modified files only.
+            repo_root = darker.black_compat.find_project_root([str(root)])
+            modified_files = git_get_modified_python_files(paths, revrange, repo_root)
+            changed_files = {
+                (repo_root / file).relative_to(root) for file in modified_files
+            }
```
akaihola, line 581: I admit the original code is confusing since it uses the words "modified" and "changed" interchangeably.

But I think this now adds further confusion by storing in
- `modified_files` the relative paths of modified files starting from the common root of given paths
- `changed_files` the relative paths of modified files starting from the Git repository root

Maybe this could be simplified to:
```suggestion
            changed_files = {
                (repo_root / file).relative_to(root)
                for file in
                git_get_modified_python_files(paths, revrange, repo_root)
            }
```

src/darker/__main__.py
```diff
@@ -572,9 +573,16 @@ def main(  # pylint: disable=too-many-locals,too-many-branches,too-many-statemen
...
         if git_is_repository(root):
-            changed_files_to_reformat = git_get_modified_python_files(
-                files_to_process, revrange, root
-            )
+            # Get the modified files only.
+            repo_root = darker.black_compat.find_project_root([str(root)])
+            changed_files = {
+                (repo_root / file).relative_to(root)
+                for file in
+                git_get_modified_python_files(paths, revrange, repo_root)
```
akaihola, line 581: Sorry for bad formatting in my suggestion. Black actually wants to have it like this:
```suggestion
                for file in git_get_modified_python_files(paths, revrange, repo_root)
```

src/darker/tests/test_main.py
```diff
@@ -929,3 +937,22 @@ def test_stdout_path_resolution(git_repo, capsys):
 
     assert result == 0
     assert capsys.readouterr().out == 'print("foo")\n'
+
+
+@pytest.mark.skipif(not sys.platform.startswith("win"), reason="Argument length limit on Windows")
```
akaihola, line 942: This line is too long, Black wants to reformat it as:
```suggestion
@pytest.mark.skipif(
    not sys.platform.startswith("win"), reason="Argument length limit on Windows"
)
```

src/darker/tests/test_main.py
```diff
@@ -929,3 +937,22 @@ def test_stdout_path_resolution(git_repo, capsys):
...
+    # on windows breaks subprocess
+    # Need to exceed 32762 characters
+    files = {}
+    path = "src"
+    for _f in range(0, 4):
+        # Of course windows limits the path length too
+        path = f"{path}/{randomword(30)}"
+
+    for _d in range(0, 210):
+       files[f"{path}/{randomword(30)}.py"] = randomword(10)
```
akaihola, line 954: Four spaces needed for indentation (by Black):
```suggestion
        files[f"{path}/{randomword(30)}.py"] = randomword(10)
```

src/darker/tests/test_main.py
```diff
@@ -929,3 +937,24 @@ def test_stdout_path_resolution(git_repo, capsys):
 
     assert result == 0
     assert capsys.readouterr().out == 'print("foo")\n'
+
+
+@pytest.mark.skipif(
+    not sys.platform.startswith("win"), reason="Argument length limit on Windows"
+)
```
akaihola, line 944: Wouldn't it be good to allow this test to be run on other platforms as well? It should pass everywhere, and if it suddenly won't, it would reveal a problem.

I know in this case we're looking at such stable environments as operating systems, and it's very unlikely that this would ever fail on Linux or macOS. But in principle, shouldn't we run all tests whenever we can?

Svenito: True. It’s not a specific windows test as such, and will still work under the other systems. I’ll remove the skip in the next commit.

akaihola:
@Svenito I will now rebase this on `master`, fix any conflicts and merge. Thanks a lot for your hard work!

PR #540, opened 2024-02-27
Drop support for EOL Python 3.7

hugovk:
Follow on from #537. Fixes #448.

PR #543, opened 2024-02-28
Format with Black 2024

hugovk:
As suggested in https://github.com/akaihola/darker/pull/540, apply Black's new style changes in a dedicated PR.


akaihola (commented):
So far I've intentionally _not_ blackened the whole code base of Darker, but instead used Darker itself to gradually update formatting to newest Black rules whenever a section of code is touched for some other reason in a PR.

This neatly results in Darker "eating its own dogfood", and acts as occasional real-life tests for Darker itself.

Opinions about this approach?

hugovk:
Yes, that sounds reasonable, feel free to close this.

PR #545, opened 2024-03-05
Filter out unmodified files from list of changed files

### LINKED ISSUE #544: Add option to ignore moved/deleted files
**Is your feature request related to a problem? Please describe.**
It should be possible to ignore moved or deleted files when determining if they should be formatted. We only care about formatting if code as been touched inside the file, not if the file has been moved or deleted.

**Describe the solution you'd like**
Add a new option to only process files that have had their content changed

**Additional context**
This could be done by adding `--diff-filter=M` to the `_git_diff_name_only` function if the command line flag is set

Svenito:
~Adds an option `--modified-only/-m` to ignore renamed files from being checked.~

Make the filter the default behaviour. Possibly add an option to include renamed files or not, depending on discussions below

Fixes #544

### DIFF
--- a/src/darker/tests/test_git.py
+++ b/src/darker/tests/test_git.py
@@ -214,16 +214,16 @@ def test_get_missing_at_revision_worktree(git_repo):
 
 
 def test_git_diff_name_only(git_repo):
-    """``_git_diff_name_only()`` only returns paths of modified files"""
+    """``_git_diff_name_only()`` includes added/modified, skips renamed/moved files."""
     git_repo.add({"a.py": "a", "b.py": "b", "c.py": "c"}, commit="Initial commit")
     first = git_repo.get_hash()
     git_repo.add({"a.py": "A", "b.dy": "B"}, commit="only a.py modified")
+    git_repo.rename("c.py", "x.py", commit="rename c.py to x.py")
     second = git_repo.get_hash()
 
     result = git._git_diff_name_only(
         first, second, {Path("a.py"), Path("c.py"), Path("Z.py")}, git_repo.root
     )
-
     assert result == {Path("a.py")}
 
 
--- a/src/darker/tests/test_git.py
+++ b/src/darker/tests/test_git.py
@@ -337,7 +337,6 @@ def test_git_get_modified_python_files(git_repo, modify_paths, paths, expect):
         _description="from master to worktree and index on branch",
         revrange="master..",
         expect={
-            "del_master.py",
             "mod_master.py",
             "mod_both.py",
             "mod_branch.py",
@@ -353,7 +352,6 @@ def test_git_get_modified_python_files(git_repo, modify_paths, paths, expect):
         ),
         revrange="master..HEAD",
         expect={
-            "del_master.py",
             "mod_master.py",
             "mod_both.py",
             "mod_branch.py",
@@ -363,7 +361,6 @@ def test_git_get_modified_python_files(git_repo, modify_paths, paths, expect):
         _description="from master to branch, excluding worktree and index",
         revrange="master..branch",
         expect={
-            "del_master.py",
             "mod_master.py",
             "mod_both.py",
             "mod_branch.py",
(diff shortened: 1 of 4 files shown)


akaihola:
Should this behavior maybe be the default? I mean, now that I think of it, if I just rename a source code file, why would I want the whole file to be reformatted in the same branch?

Svenito:
I added it because you probably don’t want to check/format files unless their content has changed. Even when slowly phasing Black into a project. 

I made it off by default because the original behaviour was to not exclude these files. If you are happy to make it default and remove the option, it’d be easier as the argument can be removed and the fit diff always filtered. If that is desired behaviour that is. 

Let me k ow what your thoughts are in this.

Timple:
For me (user) it makes sense that this is the default!

akaihola:
Thanks @Timple, and indeed @Svenito let's remove the option and hard-code this behavior as the default. I'll bump the major version number before including this in a release. Thanks!

Svenito:
I made the changes in the last commit. Also added ‘A’ to the diff filter.

akaihola:
@Svenito, could you update the unit tests? See [test failures](https://github.com/akaihola/darker/actions/runs/8185693521/job/22382555996?pr=545) in `test_git.py` and `test_main_revision.py`. Let me know if you need me to clarify how the tests work.

src/darker/tests/test_git.py
```diff
@@ -143,16 +143,17 @@ def test_get_missing_at_revision_worktree(git_repo):
...
+    renamed/moved files"""
     git_repo.add({"a.py": "a", "b.py": "b", "c.py": "c"}, commit="Initial commit")
     first = git_repo.get_hash()
     git_repo.add({"a.py": "A", "b.dy": "B"}, commit="only a.py modified")
+    git_repo.rename("c.py", "x.py", commit="rename c.py to x.py")
     second = git_repo.get_hash()
 
     result = git._git_diff_name_only(
         first, second, {Path("a.py"), Path("c.py"), Path("Z.py")}, git_repo.root
     )
```
akaihola, line 156: In test functions, I've followed the convention of separating the invocation of tested code with single blank lines above and below. This helps visually recognize the setup, test and verification parts of the test function.
```suggestion
    )

```

src/darker/tests/conftest.py
akaihola, line 1: This file can probably now be removed completely?

akaihola (commented):
akaihola/darkgraylib#38 will be included in `darkgraylib==1.1.0`. So you could rebase on `master`, bump the dependency in `setup.cfg`, and after I've made the release we can rerun the tests here.

src/darker/tests/conftest.py
Svenito: Yep. I didn’t realise it got added in again. Will remove.

akaihola:
@Svenito unfortunately you'll need to do one more rebase on `master` before the tests have a chance to pass.

Also I'll wait for at least one more backwards incompatible change to land before merging this one and bumping the major version number. So probably later in the spring we'll get this one into a release.

Thanks a lot for your effort and sorry for the delay!

Svenito:
Sounds good to me. No problem on the delay either.

Svenito:
Hi, just wanted to check if there's a rough ETA on this patch making it in?

akaihola:
> Hi, just wanted to check if there's a rough ETA on this patch making it in?

I guess it's starting to be time to bring this in. There are no significant backwards-incompatible changes to bundle with it in v3.0.0, so maybe we'll just go with this one and risk needing to bump to v4 sooner than later.

src/darker/tests/conftest.py
akaihola, line 1: @Svenito, is the empty `src/darker/tests/conftest.py` necessary?

akaihola: I removed it now. Also added a change log entry.

akaihola (approved):
Looks good to merge to me now. I don't understand why 13 CI jobs are always left unrun. But I did those lints locally, all is good.

PR #561, opened 2024-03-16
Add support for Python 3.12

akaihola:


akaihola:
The Python 3.13-dev test seems to fail on Windows when trying to compile the `regex` module, and on macOS when compiling `cffi`. So not a problem in Darker itself.

PR #565, opened 2024-03-19
Update pre-commit hook in README.rst

felixvd:
This just sets the correct git tag (`v2.0.0` instead of `2.0.0`).

Thanks for your work!

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -17,6 +17,7 @@ Removed
 
 Fixed
 -----
+- Bump-version pre-commit hook pattern: ``rev: vX.Y.Z`` instead of ``X.Y.Z``.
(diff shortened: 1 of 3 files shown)


akaihola:
Thanks @felixvd. I also added a fix for the bump version patterns so the same lines in README will be auto-updated correctly in the future.

felixvd:
Thanks for the quick review and fix-up. Looks like some tests failed, but I don't think I can re-run them to check if they're flaky.

akaihola:
> some tests failed

Ah, that's just some C extension libraries with no wheels yet on PyPI for Python 3.13-dev and compilation failing on Windows and macOS.

PR #574, opened 2024-03-28
Allow `--stdout <...> -`

### LINKED ISSUE #523: `--stdout` option throws error if dash is given indicating stdin input – breaks VSCode compatibility
**Describe the bug**
-d/--stdout option should allow "-" to indicate stdin

**To Reproduce**
1. Run  `darker -d --stdin-filename=somefile -`
2. Darker exits with error "darker.config.ConfigurationError: Either --stdin-filename=<path> or exactly one Python source file which exists on disk must be provided when using the `stdout` option"

**Expected behavior**
Darker should run and accept stdin input in the same way that it does without the trailing dash.

**Environment (please complete the following information):**
 - OS macos
 - Python version 3.9.7
 - Darker version 1.7.2
 - Black version 22.6.0

**Additional context**
I am trying to use darker with the new [VSCode black-formatter extension](https://github.com/microsoft/vscode-black-formatter/tree/main). The extension sends stdin input and seems to expect stdout, which can be accomplished by configuring it to run with `-d`. The main piece keeping it from working is that darker doesn't like the dash indicating stdin, which is added by VSCode.

I belive that [validate_stdout_src](https://github.com/akaihola/darker/blob/e96018f086383a2dcfaab6825945fbee08daca2a/src/darker/config.py#L78C6-L78C6) is likely the issue. The `validate_stdin_src` function already supports the dash.

akaihola:
Fixes #523

This makes [VSCode black-formatter extension](https://github.com/microsoft/vscode-black-formatter/tree/main) configuration easier on VSCode.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,11 @@ Added
 
 Fixed
 -----
+- A dash (``-``) is now allowed as the single source filename when using the
+  ``--stdout`` option. This makes the option compatible with the
+  `new Black extension for VSCode`__.
+
+__ https://github.com/microsoft/vscode-black-formatter
 
 
 2.1.0_ - 2024-03-27
--- a/README.rst
+++ b/README.rst
@@ -549,15 +549,21 @@ Visual Studio Code
      $ where darker
      %LocalAppData%\Programs\Python\Python36-32\Scripts\darker.exe  # possible location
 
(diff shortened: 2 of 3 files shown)


Garfounkel (approved):
Looks good to me.

PR #591, opened 2024-07-27
GH action to respect `working-directory`

### LINKED ISSUE #587: Darker github action ignores `working-directory`
- [x] Add GitHub Workflow which adds as a functional test for this

**Describe the bug**
Darker github action ignores [`working-directory` setting](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#defaultsrunworking-directory).

**To Reproduce**
Steps to reproduce the behavior: take a look at these two commits in my demo repo, [one](https://github.com/dshemetov/darker-test/pull/2) and [two](https://github.com/dshemetov/darker-test/pull/3).

**Expected behavior**
Darker CI action should use the `working-directory` setting.

**Additional context**
Vaguely related to my previous report here #522. Thanks for your help!

akaihola:
:orange_circle: Pending review is [delaying](https://github.com/users/akaihola/projects/4/views/2?pane=info&statusUpdateId=50063) Darker 3.0.0 release.

---

This patch introduces the following changes to the Darker GitHub Action:
- new `working-directory` input option (defined in `action.yml`)
- new `exitcode` and `stdout` outputs, and Python helpers for setting them
- use the `commit-range` action straight from the current Darker Git checkout
- install `pip-requirements-parser` in a separate action step instead of the entrypoint script
- cache the Python environment
- exit with code 21 if specified `working-directory` doesn't exist
- refactor `pip install` into a helper function
- only create virtualenv if it doesn't already exist
- correct GitHub URL when `version` starts with `@` (pointing to a branch or tag)
- run Darker in the specified `working-directory`
- return Darker terminal output in the `stdout` output
- check exit code and output in the `working-directory` test workflow
- update action tests to match new behavior

FIxes #587

Goes into Darker 3.0.0 since exit codes will need to change in a backwards incompatible way.

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -27,6 +27,19 @@ inputs:
       NOTE: Baseline linting has been moved to the Graylint package.
     required: false
     default: ''
+  working-directory:
+    description: >-
+      Directory to run Darker in, either absolute or relative to
+      $GITHUB_WORKSPACE. By default, Darker is run in $GITHUB_WORKSPACE.
+    required: false
+    default: '.'
+outputs:
+  exitcode:
+    description: "Exit code of Darker"
+    value: ${{ steps.darker.outputs.exitcode }}
+  stdout:
+    description: "Standard output of Darker"
+    value: ${{ steps.darker.outputs.stdout }}
 branding:
   color: "black"
   icon: "check-circle"
@@ -35,8 +48,29 @@ runs:
   steps:
     - name: Commit Range
       id: commit-range
-      uses: akaihola/darker/.github/actions/commit-range@1.7.1
+      uses: ./.github/actions/commit-range
+    - name: Set up Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: '3.x'
+    - name: Get pip cache dir
+      id: pip-cache
+      run: echo "dir=$(pip cache dir)" >> $GITHUB_OUTPUT
+      shell: bash
+    - name: Cache Darker environment
+      uses: actions/cache@v4
+      with:
+        path: |
+          ${{ github.action_path }}/.darker-env
+          ${{ steps.pip-cache.outputs.dir }}
+        key: ${{ runner.os }}-darker-${{ inputs.version }}-lint-${{ inputs.lint }}
+        save-always: true
+    - name: Install dependencies
+      run: |
+        pip install pip-requirements-parser
+      shell: bash
     - name: Run Darker
+      id: darker
       run: |
         # Exists since using github.action_path + path to main script doesn't
         # work because bash interprets the backslashes in github.action_path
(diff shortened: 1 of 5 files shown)


akaihola:
@dshemetov this one is finally ready for review and validation! The `working-directory:` should now work in the GitHub action.

akaihola:
I'm still hoping for extra eyeballs for this one. Sorry for adding quite a bit of mechanics necessary for testing this feature. Let me know if more comments or supporting documentation is required.

This may be easier to review commit by commit than the whole diff at once.

Ping @dshemetov @Svenito @dwt @MatthijsBurgh whichever one of you may be able to spare some time on reviewing.

akaihola:
Now rebased on `master` after baseline linting removal.

MatthijsBurgh:
Sorry, I don't have time for this one

akaihola:
Still looking for a reviewer for this PR. You seem to have the ball rolling @clintonsteiner (thanks for all the effort!), would you be interested to look at this change to the GitHub action as well?

akaihola:
@dshemetov I just rebased this PR to allow inclusion in the 3.0.0 release. Are you still interested in this improvement and would it be possible for you to review the changes?

action.yml
```diff
@@ -35,8 +48,29 @@ runs:
   steps:
     - name: Commit Range
       id: commit-range
-      uses: akaihola/darker/.github/actions/commit-range@1.7.1
+      uses: ./.github/actions/commit-range
+    - name: Set up Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: '3.x'
+    - name: Get pip cache dir
```
clintonsteiner, line 56: Do we want to use the built in caching included in setup python instead of this custom pip cache solution? Seems to be the way other repos are trending and seems to work better than the cache

clintonsteiner (approved):

action.yml
akaihola: That would indeed be simpler. However, `setup-python`'s `pip` caching only caches `pip`'s download cache, not the created virtualenv (see [Caching packages dependencies] in `setup-python/README.md`). What I've done here is cache the complete virtualenv.

Once the virtualenv is cached, the cache should be invalidated based on the `version` and `lint` input parameters for the action.

Also, what about using [uv] instead of `pip`? That one comes with its own caching solution (see [Caching] in `uv` docs).

I guess we'd need to benchmark all the options and see which one comes out fastest for repeated runs.

[Caching packages dependencies]: https://github.com/actions/setup-python?tab=readme-ov-file#caching-packages-dependencies
[uv]: https://docs.astral.sh/uv/
[Caching]: https://docs.astral.sh/uv/guides/integration/github/#caching

akaihola: We could just use default `pip` caching from `setup-python` for now, and have a separate issue/PR for speeding up repeated runs of the action using either `uv` and/or a different caching mechanism.

clintonsteiner: That sounds good to me
I believe either way we should get a nice speedup - and this work is done already albeit potentially not as good as a full uv implementation

If we do go that way - we should try to remove usages of actions/cache, thoughts?

akaihola: I now removed `actions/cache` and added `setup-python` `cache: 'pip'`.

I'm pushing the `actions/cache` variant in another branch just in case we want to later test differences in performance.

Issue #502, opened 2023-06-26
`pyproject.toml` options with hyphens are ignored


dwt:
**Describe the bug**
It seems that Darker ignores config options in `pyproject.toml` if they are spelled using hyphens instead of underscores.

**To Reproduce**
I have configuration for a Darker pre-commit hook like this:

```yaml
  - repo: https://github.com/akaihola/darker
    rev: 1.7.1
    hooks:
      - id: darker
        args: [ --skip-string-normalization ]

```

and a `pyproject.toml` like this:

```toml
[tool.darker]
line-length = 120
skip-string-normalization = true
```

However running `pre-commit` it did not honor the line length setting in `pyproject.toml`.

Reading the documentation and other bug reports here, it seems this ought to work.

Is this an issue in `pre-commit`, perhaps something to do with the fact that the pre-commit hook acts directly on the files that are part of the commit and therefore ignores the `pyproject.toml`?

Or is this a bug in Darker that I need to file there?

Changing the pre-commit hook to include the setting does seem to be a workaround. I.e.:

```yaml
  - repo: https://github.com/akaihola/darker
    rev: 1.7.1
    hooks:
      - id: darker
        args: [ --skip-string-normalization , --line-length=120 ]
```

**Expected behavior**
Darker should honor the config option `line-length = 120` in the pyproject.toml

**Environment (please complete the following information):**
 - OS: MacOS
 - Python version 3.10.12
 - Git version 2.39.2 (Apple Git-143)
 - Darker version 1.7.1
 - Black version black, 22.8.0 (compiled: yes)
 - other reformatter and linter versions ruff==0.0.272

dwt:
This is a continuation from https://github.com/pre-commit/pre-commit/issues/2909 where I asked first, as I assumed this might be a feature of pre-commit that was misunderstanding.

akaihola:
Thanks @dwt for the report, I was just able to reproduce this and will dig deeper now to find the cause.

akaihola:
Actually @dwt I'm not sure if I really reproduced it.

Could you add verbosity to your configuration:
```yaml
  - repo: https://github.com/akaihola/darker
    rev: 1.7.1
    hooks:
      - id: darker
        args: [ --skip-string-normalization, -vv ]
```

and run `pre-commit` again.

Do you see the `line-length` setting included in debug output for both the `# Effective configuration:` section and the `# Configuration options which differ from defaults:` section?

[...]

Issue #152, opened 2021-06-27
Add support for black's `--skip-magic-trailing-comma` flag


hauntsaninja:
Introduced in black 21.4b0 https://github.com/psf/black/blob/main/CHANGES.md#black-7

akaihola:
Thanks @hauntsaninja, I'll look into this!

PR #607, opened 2024-08-01
Remove deprecated options in `[tool.darker]`

akaihola:
These options were deprecated in Darker 2.1.1 (see #578). They can still be used as command line arguments or in the `[tool.black]` section of the configuration file.

- [x] remove the options from `DarkerConfig`
- [x] resurrect tests but modify them to assert that the options are ignored in `[tool.darker]` (or cause an error)

See also:
- #50 
- #152 
- #502

:orange_circle: Pending review is [delaying](https://github.com/users/akaihola/projects/4/views/2?pane=info&statusUpdateId=50063) Darker 3.0.0 release.

### DIFF
--- a/src/darker/tests/test_command_line.py
+++ b/src/darker/tests/test_command_line.py
@@ -262,20 +262,6 @@ def test_parse_command_line(
             " `lint =` option from your configuration file.",
         },
     ),
-    dict(
-        config={"skip_string_normalization": True},
-        expect_warn={
-            "The configuration option `skip_string_normalization` in [tool.darker] is"
-            " deprecated and will be removed in Darker 3.0."
-        },
-    ),
-    dict(
-        config={"skip_magic_trailing_comma": True},
-        expect_warn={
-            "The configuration option `skip_magic_trailing_comma` in [tool.darker] is"
-            " deprecated and will be removed in Darker 3.0."
-        },
-    ),
     dict(config={"line_length": 88}, expect_warn=set()),
     dict(config={"target_version": "py37"}, expect_warn=set()),
     dict(
@@ -285,18 +271,12 @@ def test_parse_command_line(
             "check": True,
             "isort": True,
             "lint": ["dummy"],
-            "skip_string_normalization": True,
-            "skip_magic_trailing_comma": True,
             "line_length": 88,
             "target_version": "py37",
         },
         expect_warn={
             "Baseline linting has been moved to the Graylint package. Please remove the"
             " `lint =` option from your configuration file.",
-            "The configuration option `skip_magic_trailing_comma` in [tool.darker] is"
-            " deprecated and will be removed in Darker 3.0.",
-            "The configuration option `skip_string_normalization` in [tool.darker] is"
-            " deprecated and will be removed in Darker 3.0.",
         },
     ),
 )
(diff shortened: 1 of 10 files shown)


akaihola:
@CorreyL contributed `--skip-string-normalization` in #3.

@hauntsaninja originally requested the `--skip-magic-trailing-comma` option in #152, and @talhajunaidd implemented it in #153. 

Both of those are still there as a command line option, but not supported in `pyproject.toml`'s `[tool.darker]` section any more. Put them in `[tool.black]` instead.

@dwt noted in #502 that hyphenated spellings of configuration options like `skip-magic-trailing-comma =` and `skip-string-normalization =` used to be ignored by Darker in `pyproject.toml`. Those two are not a concern of Darker's any more.

Would any of you guys be willing to take another look at this PR to complete its review?

src/darker/tests/test_command_line.py
```diff
@@ -256,20 +256,6 @@ def test_parse_command_line(
     dict(config={"check": True}, expect_warn=set()),
     dict(config={"isort": True}, expect_warn=set()),
     dict(config={"lint": ["pylint"]}, expect_warn=set()),
-    dict(
-        config={"skip_string_normalization": True},
-        expect_warn={
-            "The configuration option `skip_string_normalization` in [tool.darker] is"
-            " deprecated and will be removed in Darker 3.0."
```
dwt, line 263: I would love for the information that this is moved to `[tool.black]` now to be included here, so I don't have to research that when getting the warning.

src/darker/tests/test_command_line.py
```diff
@@ -256,20 +256,6 @@ def test_parse_command_line(
...
-        },
-    ),
-    dict(
-        config={"skip_magic_trailing_comma": True},
-        expect_warn={
-            "The configuration option `skip_magic_trailing_comma` in [tool.darker] is"
-            " deprecated and will be removed in Darker 3.0."
```
dwt, line 270: I would love for the information that this is moved to `[tool.black]` now to be included here, so I don't have to research that when getting the warning.

src/darker/tests/test_command_line.py
```diff
@@ -279,23 +265,21 @@ def test_parse_command_line(
...
-            "skip_string_normalization": True,
-            "skip_magic_trailing_comma": True,
             "line_length": 88,
             "target_version": "py37",
         },
-        expect_warn={
-            "The configuration option `skip_magic_trailing_comma` in [tool.darker] is"
-            " deprecated and will be removed in Darker 3.0.",
-            "The configuration option `skip_string_normalization` in [tool.darker] is"
-            " deprecated and will be removed in Darker 3.0.",
```
dwt, line 291: I would love for the information that this is moved to `[tool.black]` now to be included here, so I don't have to research that when getting the warning.

dwt (changes requested):
Looks good, but I would like the information about where the options move to to be include din the deprecation warnings to spare the user a research step and give him the solution immediately.

src/darker/tests/test_command_line.py
akaihola: @dwt that would have indeed been a good thing during the deprecation period.

But what now when we're actually removing these opions? Are you saying it would still be good to have Darker recognize them but give a hard error mentioning `[tool.black]`, and stop?

akaihola: @dwt I now changed it so these configuration options are parsed from the configuration file, but a descriptive error is raised if they are used. The error instructs to use the `[tool.black]` section instead.

dwt: I really like that. Thanks

PR #608, opened 2024-08-01
Remove baseline linting support

### LINKED ISSUE #466: Remove the `-L` / `--lint` option
If #465 is implemented and the linting support implementation is moved to a separate Python package, the next major release is an opportunity to drop linting support from Darker completely.

See #304 for discussion leading to this proposal.

akaihola:
Removed
-------
- **Backwards incompatible change:** Remove baseline linting support (``-L``/``--lint`` option). Use the [Graylint](/akaihola/graylint) tool instead.

Fixes #466


akaihola:
@okuuva hey since the idea to separate formatter and linter support into two tools came originally from you, would you be interested to do a quick review of this PR?

okuuva:
> @okuuva hey since the idea to separate formatter and linter support into two tools came originally from you, would you be interested to do a quick review of this PR?

Sure!

action/tests/test_main.py
```diff
@@ -2,18 +2,21 @@
...
 from types import SimpleNamespace
-from typing import Dict, Generator
+from typing import TYPE_CHECKING, Generator
 from unittest.mock import ANY, Mock, call, patch
 
 import pytest
 
+if TYPE_CHECKING:
+    from pathlib import Path
+
```
okuuva, line 19: Question: I keep seeing this pattern in some Python projects (not just in `darker`). Is there some clear benefit on conditional importing type hints? I've tried to avoid it since I ran into some problems with it in the past but maybe the situation has changes since then.

action/tests/test_main.py
```diff
@@ -98,24 +102,12 @@ def test_creates_virtualenv(tmp_path, main_patch):
...
-        run_main_env={"INPUT_LINT": "  flake8  ,  pylint  "},
-        expect=["darker[color,isort]", "flake8", "pylint"],
+        run_main_env={"INPUT_LINT": "dummy"},
+        expect=["darker[color,isort]"],
     ),
     dict(
-        run_main_env={"INPUT_LINT": "  flake8  >=  3.9.2  ,  pylint  ==  2.13.1  "},
-        expect=["darker[color,isort]", "flake8>=3.9.2", "pylint==2.13.1"],
+        run_main_env={"INPUT_LINT": "dummy,foobar"},
+        expect=["darker[color,isort]"],
```
okuuva, line 110: Question: is the point of these dummy inputs to test that using `--lint` flag doesn't break anything?

src/darker/help.py
```diff
@@ -100,12 +102,19 @@ def get_extra_instruction(dependency: str) -> str:
...
+    Baseline linting support has been moved to the Graylint package. Please install
+    and run it using:
+
+        pip install graylint
+        graylint -L CMD <directory>
+        graylint --lint CMD <directory>
+
+    See https://pypi.org/project/graylint for more information."
+
+    """,
```
okuuva, line 117: Comment: I was today years old when I learned about `textwrap.dedent`. Coolness!

okuuva (approved):
Just one question about opinionated(?) thing and another for clarification. All in all LGTM!

action/tests/test_main.py
akaihola: Linters (e.g. `ruff`) insist on this. On one occasion putting the import outside the if clause would have caused an import loop in my code. And I guess in some cases unneeded imports might be avoided (in case the user never hits some particular functionality).

PR #612, opened 2024-08-09
Work around `Path.resolve` bug in Windows py38/py39

akaihola:
In some PRs, tests started failing on Python 3.8 and 3.9 when run in Windows. This was due to a Python bug, see python/cpython#82852. This PR adds a special case for those environments until we drop support for Python 3.9.

### DIFF
--- a/src/darker/__main__.py
+++ b/src/darker/__main__.py
@@ -52,7 +52,7 @@ from darkgraylib.git import (
 from darkgraylib.highlighting import colorize, should_use_color
 from darkgraylib.log import setup_logging
 from darkgraylib.main import resolve_paths
-from darkgraylib.utils import GIT_DATEFORMAT, DiffChunk, TextDocument
+from darkgraylib.utils import GIT_DATEFORMAT, WINDOWS, DiffChunk, TextDocument
 from graylint.linting import run_linters
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


src/darker/__main__.py
```diff
@@ -561,9 +561,18 @@ def main(  # noqa: C901,PLR0912,PLR0915
...
             raise FileNotFoundError(msg)
 
+    common_root_ = (
+        # On Windows, Python <= 3.9 requires the `filter_python_files` `root` argument
+        # to be an absolute path. Remove this after dropping support for Python 3.9.
+        # See https://github.com/python/cpython/issues/82852
+        common_root.resolve()
+        if WINDOWS and sys.version_info < (3, 10)
+        else common_root
+    )
```
DeinAlptraum, line 571: According to the issue you linked, this bug also affects Python 3.10

akaihola: For some reason the unit tests only failed on Python 3.8 and 3.9 in the Windows runners of the Github workflow. I'll look into this, thanks for the heads up!

akaihola: @DeinAlptraum I checked this; python/cpython#82852 was completed on August 7th, 2021, and Python 3.10 was released, including that bugfix, on October 4th, 2021.

So the bug was in the developing version 3.10 but was fixed before the release.

PR #738, opened 2024-09-15
Black as a plugin

akaihola:
- [x] Merge #568 
- [x] Rebase on `master`

Enables:
- #744 

(see [many-formatters](https://github.com/users/akaihola/projects/4/views/5) sub-project).


akaihola:
Hi @clintonsteiner, this is the PR leading up to #744 (ruff support). You already reviewed the draft PR #744 which also contains these commits. I did a bit of clean-up and rebased this one again, and I believe this is now ready to be reviewed and merged. After this one I'd like to review and merge #759 (drop Black dependency) as well.

src/darker/__main__.py
```diff
@@ -1,4 +1,4 @@
-"""Darker - apply black reformatting to only areas edited since the last commit"""
+"""Darker - re-format only code areas edited since the last commit."""
```
clintonsteiner, line 1: nit: could just be format which I think is marginally cleaner

src/darker/formatters/black_formatter.py
```diff
@@ -0,0 +1,222 @@
...
+            )
+
+    def _read_cli_args(self, args: Namespace) -> None:
+        if args.config:
+            self.config["config"] = args.config
+        if getattr(args, "line_length", None):
+            self.config["line_length"] = args.line_length
+        if getattr(args, "target_version", None):
+            self.config["target_version"] = {args.target_version}
+        if getattr(args, "skip_string_normalization", None) is not None:
```
clintonsteiner, line 142: nit: can remove is not None and just do if as used in the above

clintonsteiner (approved):

src/darker/__main__.py
akaihola: Re-wrote that as
>Darker - format only code areas modified between Git revisions.

src/darker/formatters/black_formatter.py
akaihola: But Darker has the `--no-skip-string-normalization` option, and if it's specified, we want to add `skip_string_normalization=False` to the configuration.

Issue #465, opened 2023-02-11
Move linting related code to a new `graylint` package


akaihola:
Running linters doesn't really fit together along with Darker's original purpose of executing code reformatters. Reformatters make local modifications to code, while linters only report rule violations and may detect flaws in one part of the code in relation to a different part (e.g. undefined identifiers).

Eventually we should remove `-L` / `--lint` from Darker completely, but before that we'd like to separate out linting functionality into a dedicated tool (named e.g. `graylint`) and move all related code there. Code needed by both Darker and that package should go into a third library support package (named e.g. `libdarkgray`). Darker can deprecate but still keep the `-L` / `--lint` option, and depend on `graylint` to implement it.

Alternatives considered:
- all code in the Darker repo, but in three different Python packages which are published separately on PyPI
- rename the Darker package to something more generic, and provide two console entry points `darker` and `graylint` in it
- duplicate all common code in both Darker and Graylint
- keep linting functionality in Darker, rename Darker to something more generic, add `--black` / `--no-black` and make `--no-black` the default, and re-create Darker as a thin wrapper with `--black` as the default

See #304 for discussion leading to this proposal.

PR #744, opened 2024-09-17
Add support for Ruff as a code-formatter

### LINKED ISSUE #521: Support `ruff format` as a formatter
Any plans to begin using [`ruff format`](https://docs.astral.sh/ruff/formatter/) instead of black for code formatting?
Ruff is designed to be a drop in replacement for black and should provide a nice performance boost

---

_@akaihola:_

>Some pointers to get started with implementation:
>- [many-formatters](/users/akaihola/projects/4/views/5) sub-project
>- #251
>- #304
>- #51
>- #309
>- #266

akaihola:
Fixes #521

- [x] Review #568
- [x] Merge #568
- [x] Review #738
- [x] Merge #738
- [x] Review #761
- [x] Merge #761
- [x] Rebase on `master`
- [x] Review #759
- [x] Merge #759
- [x] Rebase on `master` 
- [x] Review #792
- [x] Merge #792
- [x] Rebase on `master`

### DIFF
--- a/src/darker/__main__.py
+++ b/src/darker/__main__.py
@@ -293,7 +293,7 @@ def _maybe_reformat_single_file(
     if glob_any(relpath_in_rev2, exclude):
         # File was excluded by Black configuration, don't reformat
         return fstringified
-    return formatter.run(fstringified)
+    return formatter.run(fstringified, relpath_in_rev2)
 
 
 def _drop_changes_on_unedited_lines(
--- a/src/darker/formatters/black_formatter.py
+++ b/src/darker/formatters/black_formatter.py
@@ -117,10 +117,15 @@ class BlackFormatter(BaseFormatter, HasConfig[BlackCompatibleConfig]):
         if "target_version" in raw_config:
             target_version = raw_config["target_version"]
             if isinstance(target_version, str):
-                self.config["target_version"] = target_version
+                self.config["target_version"] = (
+                    int(target_version[2]),
+                    int(target_version[3:]),
+                )
             elif isinstance(target_version, list):
-                # Convert TOML list to a Python set
-                self.config["target_version"] = set(target_version)
+                # Convert TOML list to a Python set of int-tuples
+                self.config["target_version"] = {
+                    (int(v[2]), int(v[3:])) for v in target_version
+                }
             else:
                 message = (
                     f"Invalid target-version = {target_version!r} in {config_path}"
@@ -183,7 +188,10 @@ class BlackFormatter(BaseFormatter, HasConfig[BlackCompatibleConfig]):
         if "line_length" in self.config:
             mode["line_length"] = self.config["line_length"]
         if "target_version" in self.config:
-            all_target_versions = {tgt_v.name.lower(): tgt_v for tgt_v in TargetVersion}
+            all_target_versions = {
+                (int(tgt_v.name[2]), int(tgt_v.name[3:])): tgt_v
+                for tgt_v in TargetVersion
+            }
             target_versions_in = validate_target_versions(
                 self.config["target_version"], all_target_versions
             )
--- a/src/darker/formatters/black_formatter.py
+++ b/src/darker/formatters/black_formatter.py
@@ -57,8 +57,6 @@ if TYPE_CHECKING:
     from black import FileMode as Mode
     from black import TargetVersion
 
-    from darker.formatters.formatter_config import BlackConfig
-
 
 logger = logging.getLogger(__name__)
 
--- a/src/darker/formatters/black_formatter.py
+++ b/src/darker/formatters/black_formatter.py
@@ -13,7 +13,7 @@ The first line will be reformatted by Black, and the second left intact::
     ...     ]
     ... )
 
-First, :func:`run_black` uses Black to reformat the contents of a given file.
+First, `BlackFormatter.run` uses Black to reformat the contents of a given file.
 Reformatted lines are returned e.g.::
 
     >>> from darker.formatters.black_formatter import BlackFormatter
@@ -40,7 +40,12 @@ import logging
 from typing import TYPE_CHECKING, TypedDict
 
 from darker.files import find_pyproject_toml
-from darker.formatters.base_formatter import BaseFormatter
+from darker.formatters.base_formatter import BaseFormatter, HasConfig
+from darker.formatters.formatter_config import (
+    BlackCompatibleConfig,
+    read_black_compatible_cli_args,
+    validate_target_versions,
+)
 from darkgraylib.config import ConfigurationError
 from darkgraylib.utils import TextDocument
 
@@ -68,12 +73,10 @@ class BlackModeAttributes(TypedDict, total=False):
     preview: bool
 
 
-class BlackFormatter(BaseFormatter):
+class BlackFormatter(BaseFormatter, HasConfig[BlackCompatibleConfig]):
     """Black code formatter plugin interface."""
 
-    def __init__(self) -> None:  # pylint: disable=super-init-not-called
-        """Initialize the Black code re-formatter plugin."""
-        self.config: BlackConfig = {}
+    config: BlackCompatibleConfig  # type: ignore[assignment]
 
     name = "black"
 
@@ -135,18 +138,7 @@ class BlackFormatter(BaseFormatter):
             )
 
     def _read_cli_args(self, args: Namespace) -> None:
-        if args.config:
-            self.config["config"] = args.config
-        if getattr(args, "line_length", None):
-            self.config["line_length"] = args.line_length
-        if getattr(args, "target_version", None):
-            self.config["target_version"] = {args.target_version}
-        if getattr(args, "skip_string_normalization", None) is not None:
-            self.config["skip_string_normalization"] = args.skip_string_normalization
-        if getattr(args, "skip_magic_trailing_comma", None) is not None:
-            self.config["skip_magic_trailing_comma"] = args.skip_magic_trailing_comma
-        if getattr(args, "preview", None):
-            self.config["preview"] = args.preview
+        return read_black_compatible_cli_args(args, self.config)
 
     def run(self, content: TextDocument) -> TextDocument:
         """Run the Black code re-formatter for the Python source code given as a string.
@@ -191,15 +183,10 @@ class BlackFormatter(BaseFormatter):
         if "line_length" in self.config:
             mode["line_length"] = self.config["line_length"]
         if "target_version" in self.config:
-            if isinstance(self.config["target_version"], set):
-                target_versions_in = self.config["target_version"]
-            else:
-                target_versions_in = {self.config["target_version"]}
             all_target_versions = {tgt_v.name.lower(): tgt_v for tgt_v in TargetVersion}
-            bad_target_versions = target_versions_in - set(all_target_versions)
-            if bad_target_versions:
-                message = f"Invalid target version(s) {bad_target_versions}"
-                raise ConfigurationError(message)
+            target_versions_in = validate_target_versions(
+                self.config["target_version"], all_target_versions
+            )
             mode["target_versions"] = {
                 all_target_versions[n] for n in target_versions_in
             }
--- a/src/darker/formatters/black_formatter.py
+++ b/src/darker/formatters/black_formatter.py
@@ -17,7 +17,7 @@ First, `BlackFormatter.run` uses Black to reformat the contents of a given file.
 Reformatted lines are returned e.g.::
 
     >>> from darker.formatters.black_formatter import BlackFormatter
-    >>> dst = BlackFormatter().run(src_content)
+    >>> dst = BlackFormatter().run(src_content, src)
     >>> dst.lines
     ('for i in range(5):', '    print(i)', 'print("done")')
 
@@ -51,6 +51,7 @@ from darkgraylib.utils import TextDocument
 
 if TYPE_CHECKING:
     from argparse import Namespace
+    from pathlib import Path
     from typing import Pattern
 
     from black import FileMode as Mode
@@ -145,10 +146,14 @@ class BlackFormatter(BaseFormatter, HasConfig[BlackCompatibleConfig]):
     def _read_cli_args(self, args: Namespace) -> None:
         return read_black_compatible_cli_args(args, self.config)
 
-    def run(self, content: TextDocument) -> TextDocument:
+    def run(
+        self, content: TextDocument, path_from_cwd: Path  # noqa: ARG002
+    ) -> TextDocument:
         """Run the Black code re-formatter for the Python source code given as a string.
 
         :param content: The source code
+        :param path_from_cwd: The path to the source code file being reformatted, either
+                              absolute or relative to the current working directory
         :return: The reformatted content
 
         """
--- a/src/darker/formatters/black_formatter.py
+++ b/src/darker/formatters/black_formatter.py
@@ -80,6 +80,7 @@ class BlackFormatter(BaseFormatter, HasConfig[BlackCompatibleConfig]):
     config: BlackCompatibleConfig  # type: ignore[assignment]
 
     name = "black"
+    config_section = "tool.black"
 
     def read_config(self, src: tuple[str, ...], args: Namespace) -> None:
         """Read Black configuration from ``pyproject.toml``.
--- a/src/darker/formatters/formatter_config.py
+++ b/src/darker/formatters/formatter_config.py
@@ -2,8 +2,6 @@
 
 from __future__ import annotations
 
-import re
-from decimal import Decimal
 from typing import TYPE_CHECKING, Iterable, Pattern, TypedDict
 
 from darkgraylib.config import ConfigurationError
@@ -17,11 +15,12 @@ class FormatterConfig(TypedDict):
 
 
 def validate_target_versions(
-    value: str | set[str], valid_target_versions: Iterable[str]
-) -> set[str]:
+    value: tuple[int, int] | set[tuple[int, int]],
+    valid_target_versions: Iterable[tuple[int, int]],
+) -> set[tuple[int, int]]:
     """Validate the target-version configuration option value."""
-    target_versions_in = {value} if isinstance(value, str) else value
-    if not isinstance(value, (str, set)):
+    target_versions_in = value if isinstance(value, set) else {value}
+    if not isinstance(value, (tuple, set)):
         message = f"Invalid target version(s) {value!r}"  # type: ignore[unreachable]
         raise ConfigurationError(message)
     bad_target_versions = target_versions_in - set(valid_target_versions)
@@ -31,16 +30,6 @@ def validate_target_versions(
     return target_versions_in
 
 
-get_version_num = re.compile(r"\d+").search
-
-
-def get_minimum_target_version(target_versions: set[str]) -> str:
-    """Get the minimum target version from a set of target versions."""
-    nums_and_tgts = ((get_version_num(tgt), tgt) for tgt in target_versions)
-    matches = ((Decimal(match.group()), tgt) for match, tgt in nums_and_tgts if match)
-    return min(matches)[1]
-
-
 class BlackCompatibleConfig(FormatterConfig, total=False):
     """Type definition for configuration dictionaries of Black compatible formatters."""
 
@@ -48,7 +37,7 @@ class BlackCompatibleConfig(FormatterConfig, total=False):
     exclude: Pattern[str]
     extend_exclude: Pattern[str] | None
     force_exclude: Pattern[str] | None
-    target_version: str | set[str]
+    target_version: tuple[int, int] | set[tuple[int, int]]
     line_length: int
     skip_string_normalization: bool
     skip_magic_trailing_comma: bool
@@ -64,7 +53,9 @@ def read_black_compatible_cli_args(
     if getattr(args, "line_length", None):
         config["line_length"] = args.line_length
     if getattr(args, "target_version", None):
-        config["target_version"] = {args.target_version}
+        config["target_version"] = {
+            (int(args.target_version[2]), int(args.target_version[3:]))
+        }
     if getattr(args, "skip_string_normalization", None) is not None:
         config["skip_string_normalization"] = args.skip_string_normalization
     if getattr(args, "skip_magic_trailing_comma", None) is not None:
--- a/src/darker/formatters/formatter_config.py
+++ b/src/darker/formatters/formatter_config.py
@@ -2,15 +2,47 @@
 
 from __future__ import annotations
 
-from typing import Pattern, TypedDict
+import re
+from decimal import Decimal
+from typing import TYPE_CHECKING, Iterable, Pattern, TypedDict
+
+from darkgraylib.config import ConfigurationError
+
+if TYPE_CHECKING:
+    from argparse import Namespace
 
 
 class FormatterConfig(TypedDict):
     """Base class for code re-formatter configuration."""
 
 
-class BlackConfig(FormatterConfig, total=False):
-    """Type definition for Black configuration dictionaries."""
+def validate_target_versions(
+    value: str | set[str], valid_target_versions: Iterable[str]
+) -> set[str]:
+    """Validate the target-version configuration option value."""
+    target_versions_in = {value} if isinstance(value, str) else value
+    if not isinstance(value, (str, set)):
+        message = f"Invalid target version(s) {value!r}"  # type: ignore[unreachable]
+        raise ConfigurationError(message)
+    bad_target_versions = target_versions_in - set(valid_target_versions)
+    if bad_target_versions:
+        message = f"Invalid target version(s) {bad_target_versions}"
+        raise ConfigurationError(message)
+    return target_versions_in
+
+
+get_version_num = re.compile(r"\d+").search
+
+
+def get_minimum_target_version(target_versions: set[str]) -> str:
+    """Get the minimum target version from a set of target versions."""
+    nums_and_tgts = ((get_version_num(tgt), tgt) for tgt in target_versions)
+    matches = ((Decimal(match.group()), tgt) for match, tgt in nums_and_tgts if match)
+    return min(matches)[1]
+
+
+class BlackCompatibleConfig(FormatterConfig, total=False):
+    """Type definition for configuration dictionaries of Black compatible formatters."""
 
     config: str
     exclude: Pattern[str]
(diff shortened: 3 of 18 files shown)


akaihola:
FYI @clintonsteiner @mrfroggg @Svenito support for Ruff as a code re-formatter is soon landing. This branch basically works already, we just need to add tests and documentation and clean up a bit. After that, code reviews would be highly appreciated! I'll invite you as collaborators on the repository just in case you might have time to review.

clintonsteiner:
Sounds great! Let me know what I can do to help

clintonsteiner:
I'd be happy to review, add tests, or add docs. Thank you!

clintonsteiner:
I have some changes - would like to help.
Would you prefer I make a pull request on top of this branch?

akaihola:
Hi @clintonsteiner, thanks for your reply! I just rebased this branch on `master`, feel free to make your pull requests on top of this branch.

Note that there's a stack of branches sitting on top of each other, and I'm planning to merge them one by one and rebase the latter ones on `master` after merging:
- #568
  - branch [`formatter-option'](/akaihola/darker/tree/formatter-option)
- #738
  - branch [`black-as-plugin`](/akaihola/darker/tree/black-as-plugin)
- #744 (this PR)

If some of your fixes are about #568 or #738, it's ok for me if you still push them to a PR on top of #744. I can then do the clean-up work and move them up to the correct branch.

But reviewing #568 and #738 would certainly also help!

src/darker/help.py
```diff
@@ -4,6 +4,8 @@
```
clintonsteiner, line 4: nit: spacing

src/darker/tests/test_command_line.py
```diff
@@ -529,14 +546,101 @@ def test_black_options(monkeypatch, tmpdir, git_repo, options, expect):
...
+        expect_opts=[
+            "--line-length=81",
+            "--target-version=py39",
+            '--config=format.quote-style="preserve"',
+        ],
+    ),
+    dict(options=["--preview"], expect_opts=["--preview"]),
+    expect_opts=[],
+)
+def test_ruff_options(monkeypatch, tmpdir, git_repo, capsys, options, expect_opts):
```
clintonsteiner, line 613: capsys truly not used from what I can tell

src/darker/tests/test_command_line.py
```diff
@@ -529,14 +546,101 @@ def test_black_options(monkeypatch, tmpdir, git_repo, options, expect):
...
+    """Ruff options from the command line are passed correctly to Ruff"""
+    monkeypatch.chdir(tmpdir)
+    (tmpdir / "pyproject.toml").write("[tool.ruff]\n")
```
clintonsteiner, line 616: os.path.join(tmpdir, "pyproject.toml") os agnostic

src/darker/formatters/ruff_formatter.py
```diff
@@ -0,0 +1,198 @@
...
+from darker.formatters.formatter_config import (
+    BlackCompatibleConfig,
+    get_minimum_target_version,
+    read_black_compatible_cli_args,
+    validate_target_versions,
+)
+from darkgraylib.config import ConfigurationError
+from darkgraylib.utils import TextDocument
+
+if sys.version_info >= (3, 11):
```
clintonsteiner, line 56: Comment could be helpful that this means python 3.11, I had to look it up

src/darker/formatters/ruff_formatter.py
```diff
@@ -0,0 +1,198 @@
...
+    """
+    cmdline = "ruff config target-version"
+    output = run(  # noqa: S603  # nosec
+        cmdline.split(), stdout=PIPE, check=True, text=True
+    ).stdout
+    type_lines = [line for line in output.splitlines() if line.startswith('Type: "py')]
+    if not type_lines:
+        message = f"`{cmdline}` returned no target versions on a 'Type: \"py...' line"
+        raise ConfigurationError(message)
+    quoted_targets = type_lines[0][6:].split(" | ")
```
clintonsteiner, line 167: Magic number 6 - what does that mean here

src/darker/formatters/ruff_formatter.py
```diff
@@ -0,0 +1,198 @@
...
+    if any(tgt_ver[0] != '"' or tgt_ver[-1] != '"' for tgt_ver in quoted_targets):
+        message = f"`{cmdline}` returned invalid target versions {type_lines[0]!r}"
```
clintonsteiner, line 169: !r at the end I don't understand what it does

src/darker/formatters/ruff_formatter.py
```diff
@@ -0,0 +1,198 @@
...
+    cmdline = "ruff config target-version"
+    output = run(  # noqa: S603  # nosec
+        cmdline.split(), stdout=PIPE, check=True, text=True
+    ).stdout
+    type_lines = [line for line in output.splitlines() if line.startswith('Type: "py')]
+    if not type_lines:
+        message = f"`{cmdline}` returned no target versions on a 'Type: \"py...' line"
+        raise ConfigurationError(message)
+    quoted_targets = type_lines[0][6:].split(" | ")
+    if any(tgt_ver[0] != '"' or tgt_ver[-1] != '"' for tgt_ver in quoted_targets):
```
clintonsteiner, line 168: if '"' not in [tgt_ver[0], tgt_ver[-1]] for tgt_ver in quoted_targets reads cleaner to me
Comment would be appreciated here as well

src/darker/formatters/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+from importlib.metadata import EntryPoint, entry_points
+from typing import cast
+
+from darker.formatters.base_formatter import BaseFormatter
+
+ENTRY_POINT_GROUP = "darker.formatter"
+
+
+def get_formatter_entry_points(name: str | None = None) -> tuple[EntryPoint, ...]:
+    """Get the entry points of all built-in code re-formatter plugins."""
```
clintonsteiner, line 15: Sorry probably missing something, but are we registereing all entry points for all currently installed formatters?
Would this register black and ruff?
Are we not supporting that below 3.10?

src/darker/formatters/black_formatter.py
```diff
@@ -0,0 +1,205 @@
...
+
+        :param src: The source code files and directories to be processed by Darker
+        :param args: Command line arguments
+
+        """
+        config_path = args.config or find_pyproject_toml(src)
+        if config_path:
+            self._read_config_file(config_path)
+        self._read_cli_args(args)
+
```
clintonsteiner, line 99: Helper method would help cleanup lines 100->125
def _add_to_config_if_exists(self, raw_config, keys):
    for name in keys:
        if name in raw_config:
            self.config[name] = raw_config[name]

src/darker/__main__.py
```diff
@@ -213,9 +213,10 @@ def _blacken_and_flynt_single_file(
         len(fstringified.lines),
         "some" if has_fstring_changes else "no",
     )
-    # 3. run black on the isorted and fstringified contents of each edited to-file
-    formatted = _maybe_blacken_single_file(
-        relative_path_in_rev2, exclude.black, fstringified, black_config
+    # 3. run the code re-formatter on the isorted and fstringified contents of each
+    #    edited to-file
```
clintonsteiner, line 217: Could we add a debug log showing which formatter we are using

src/darker/files.py
```diff
@@ -25,3 +41,49 @@ def find_pyproject_toml(path_search_start: Tuple[str, ...]) -> Optional[str]:
...
+    :param formatter: The code re-formatter which provides the configuration containing
+                      the exclude options
+    :return: Paths of files which should be reformatted according to
+             ``black_config``, relative to ``root``.
+
+    """
+    sig = inspect.signature(gen_python_files)
+    # those two exist and are required in black>=21.7b1.dev9
+    kwargs = {"verbose": False, "quiet": False} if "verbose" in sig.parameters else {}
+    # `gitignore=` was replaced with `gitignore_dict=` in black==22.10.1.dev19+gffaaf48
```
clintonsteiner, line 64: Shouldn't need to iterate across params
Just if "gitignore" in sig.parameters:
kwargs["gitignore"] = None

src/darker/files.py
```diff
@@ -25,3 +41,49 @@ def find_pyproject_toml(path_search_start: Tuple[str, ...]) -> Optional[str]:
...
+    for param in sig.parameters:
+        if param == "gitignore":
+            kwargs[param] = None  # type: ignore[assignment]
+        elif param == "gitignore_dict":
+            kwargs[param] = {}  # type: ignore[assignment]
+    absolute_paths = {p.resolve() for p in paths}
```
clintonsteiner, line 70: Iterate once with 
directories, files = {}, {}
for p in paths:
    path = p.resolve()
    if path.is_dir():
        directories.add(path)
    else:
        files.add(path)

src/darker/files.py
```diff
@@ -25,3 +41,49 @@ def find_pyproject_toml(path_search_start: Tuple[str, ...]) -> Optional[str]:
...
+    directories = {p for p in absolute_paths if p.is_dir()}
+    files = {p for p in absolute_paths if p not in directories}
+    files_from_directories = set(
```
clintonsteiner, line 73: Comment could be helpful explaining recursing dirs to get python files

src/darker/files.py
```diff
@@ -25,3 +41,49 @@ def find_pyproject_toml(path_search_start: Tuple[str, ...]) -> Optional[str]:
...
+            root,
+            include=DEFAULT_INCLUDE_RE,
+            exclude=formatter.get_exclude(DEFAULT_EXCLUDE_RE),
+            extend_exclude=formatter.get_extend_exclude(),
+            force_exclude=formatter.get_force_exclude(),
+            report=Report(),
+            **kwargs,  # type: ignore[arg-type]
+        )
+    )
+    return {p.resolve().relative_to(root) for p in files_from_directories | files}
```
clintonsteiner, line 85: More readable could be
return {p.resolve().relative_to(root) for p in  files.union(files_from_directories)}

src/darker/formatters/black_formatter.py
```diff
@@ -0,0 +1,205 @@
...
+        if "skip_magic_trailing_comma" in self.config:
+            mode["magic_trailing_comma"] = not self.config["skip_magic_trailing_comma"]
+        if "skip_string_normalization" in self.config:
+            # The ``black`` command line argument is
+            # ``--skip-string-normalization``, but the parameter for
+            # ``black.Mode`` needs to be the opposite boolean of
+            # ``skip-string-normalization``, hence the inverse boolean
+            mode["string_normalization"] = not self.config["skip_string_normalization"]
+        if "preview" in self.config:
+            mode["preview"] = self.config["preview"]
```
clintonsteiner, line 172: IMO above should be moved out of run function since just related to setting up config of the class

src/darker/formatters/formatter_config.py
```diff
@@ -0,0 +1,73 @@
...
+from typing import TYPE_CHECKING, Iterable, Pattern, TypedDict
+
+from darkgraylib.config import ConfigurationError
+
+if TYPE_CHECKING:
+    from argparse import Namespace
+
+
+class FormatterConfig(TypedDict):
+    """Base class for code re-formatter configuration."""
```
clintonsteiner, line 16: If this is the only place this is used can probably jsut remove it

src/darker/formatters/formatter_config.py
```diff
@@ -0,0 +1,73 @@
...
+        message = f"Invalid target version(s) {value!r}"  # type: ignore[unreachable]
+        raise ConfigurationError(message)
+    bad_target_versions = target_versions_in - set(valid_target_versions)
+    if bad_target_versions:
+        message = f"Invalid target version(s) {bad_target_versions}"
+        raise ConfigurationError(message)
+    return target_versions_in
+
+
+get_version_num = re.compile(r"\d+").search
```
clintonsteiner, line 34: Move below line 37

src/darker/formatters/formatter_config.py
```diff
@@ -0,0 +1,73 @@
...
+    args: Namespace, config: BlackCompatibleConfig
+) -> None:
+    """Read Black-compatible configuration from command line arguments."""
+    if args.config:
+        config["config"] = args.config
+    if getattr(args, "line_length", None):
+        config["line_length"] = args.line_length
+    if getattr(args, "target_version", None):
+        config["target_version"] = {args.target_version}
+    if getattr(args, "skip_string_normalization", None) is not None:
```
clintonsteiner, line 68: nit: can remove is not None

src/darker/formatters/formatter_config.py
```diff
@@ -0,0 +1,73 @@
...
+    line_length: int
+    skip_string_normalization: bool
+    skip_magic_trailing_comma: bool
+    preview: bool
+
+
+def read_black_compatible_cli_args(
+    args: Namespace, config: BlackCompatibleConfig
+) -> None:
+    """Read Black-compatible configuration from command line arguments."""
```
clintonsteiner, line 61: Can simplify with helper function for below if you wanted

clintonsteiner (commented):
Left plenty of reviews, sorry just trying to be helpful
Feel free to ignore them as you disagree with them :)

clintonsteiner (approved):

clintonsteiner:
Left a review - but not quite following this project yet
Tried to be helpful but please feel free to disregard as you see fit

zmeir:
Looking forward to using this! 😃 

Small question: will this PR support `ruff check --fix` as well? or just `ruff format`?

akaihola:
> Small question: will this PR support `ruff check --fix` as well? or just `ruff format`?

We've recently split out baseline linting support from Darker into the separate [Graylint] package (see #304, #465, #466 and #608).

Graylint is already compatible with [ruff check], and you can simply run
```bash
pip install graylint                           # you can now display linting errors from ruff caused by:
graylint --lint="ruff check" .                 # - uncommitted changes
graylint --revision=main... -L "ruff check" .  # - changes in current feature branch (-L = --lint)
graylint -r rev1..rev2 -L "ruff check" .       # - changes between branches/tags (-r = --revision)

```
Graylint [configuration] is very similar to Darker's, and they share a lot of code through [Darkgraylib].

[Graylint]: /akaihola/Graylint
[ruff check]: https://docs.astral.sh/ruff/linter/
[configuration]: /akaihola/graylint/blob/d84d790a23e6e7289ff9a70cf9b5aaefaa2f87fa/README.rst#customizing-graylint-and-linter-behavior
[Darkgraylib]: /akaihola/darkgraylib

src/darker/help.py
```diff
@@ -146,3 +148,8 @@ def get_extra_instruction(dependency: str) -> str:
 )
 
 WORKERS = "How many parallel workers to allow, or `0` for one per core [default: 1]"
+
+FORMATTER = (
+    f"[{'|'.join(f.name for f in entry_points(group="darker.formatter"))}] Formatter"
```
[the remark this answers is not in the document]
akaihola: Now uses a helper:
```python
FORMATTER = (
    f"[{'|'.join(get_formatter_names())}] Formatter"
    " to use for reformatting code. [default: black]"
)
```

src/darker/help.py
akaihola: Sorry, I can't figure out what your diff is – GitHub won't expand it to show more context. Is it about empty lines within the import section?

Here's how `isort` and `black` have sorted and grouped the imports currently:
```python
"""Help and usage instruction texts used for the command line parser"""

from textwrap import dedent

from black import TargetVersion

from darker.formatters import get_formatter_names


def get_extra_instruction(dependency: str) -> str:
```

As I understand it, the empty lines between imports separate them into three sections:
- standard library (`textwrap`),
- 3rd party packages (`black`), and
- "1st party" imports from the package itself (`darker`).

(I tend to instruct linters to consider closely related packages like `darkgraylib` also first party.)

src/darker/tests/test_command_line.py
akaihola: True!

src/darker/tests/test_command_line.py
akaihola: I was under the impression that `py.path.local() / "subpath"` using the Pytest [tmpdir] fixture is also OS agnostic (see [Working with Paths] in Pytest `py.path` documentation). I think I've used it successfully elsewhere without Windows issues. Any gotchas I'm not aware of?

But, this is a good reminder to follow Pytest's [recommendation] here, too, to migrate to the [tmp_path] fixture which returns a [pathlib.Path]! I'm pretty sure `pathlib.Path() / "subpath"` is OS agnostic (see [Operators] in `pathlib` documentation). 

[tmpdir]: https://docs.pytest.org/en/6.2.x/tmpdir.html#the-tmpdir-fixture
[Working with Paths]: https://py.readthedocs.io/en/latest/path.html#working-with-paths
[recommendation]: https://py.readthedocs.io/en/latest/path.html#py-path
[tmp_path]: https://docs.pytest.org/en/6.2.x/tmpdir.html#the-tmp-path-fixture
[pathlib.Path]: https://docs.python.org/3/library/pathlib.html#pathlib.Path
[Operators]: https://docs.python.org/3/library/pathlib.html#operators

src/darker/formatters/black_formatter.py
akaihola: Setting up the config of the formatter class is done in `_read_config_file()`.

This code is about preparing the options for the call to `black.format_str(contents_for_black, mode=Mode(**mode))` which are encapsulated inside the `Mode` object.

But true nevertheless – preparing the data for the `Mode` object can indeed be split into a helper method.

src/darker/formatters/formatter_config.py
akaihola: I'm actually reworking the target version number handling, so this will go away altogether.

src/darker/formatters/formatter_config.py
akaihola: Actually we can't. If the value is `False`, we want to add it to `config`.

src/darker/formatters/formatter_config.py
akaihola: The whole `FormatterConfig` class?

I'm using it for type checking. All configuration classes for different formatters must subclass this one, and Mypy can then check the types correctly.

I did consider using [class protocols] instead of subclassing. But this is such a simple class hierarchy that I'm not sure if it pays off.

[class protocols]: https://docs.python.org/3/library/typing.html#typing.Protocol

src/darker/formatters/formatter_config.py
akaihola: It's a matter of taste I guess. I find often than being clever in this kind of code can lead to complications in static type checking. Since it's still just 7 options, I think I prefer to keep the code a bit repetitive but simple.

src/darker/formatters/black_formatter.py
akaihola: It's neat, but we then lose static type checking. Running Mypy on such code will give us
```python
error: TypedDict key must be a string literal; expected one of (
    "config", "exclude", "extend_exclude", "force_exclude", "target_version", ...
)  [literal-required]
```

src/darker/formatters/ruff_formatter.py
akaihola: Replaced with `len('Type: ')`.

zmeir:
> > Small question: will this PR support `ruff check --fix` as well? or just `ruff format`?
> 
> 
> 
> We've recently split out baseline linting support from Darker into the separate [Graylint] package (see #304, #465, #466 and #608).

That's very cool - I'll have to try it out 😃 

However, I was referring specifically to the `--fix` option of `ruff check`, which not only reports linting errors but also fixes them - which means modifying the code.  
If I understand the function of Graylint correctly, that's outside  its scope of operation.

From #465:
> Reformatters make local modifications to code, while linters only report rule violations and may detect flaws in one part of the code in relation to a different part

Following this definition, it seems like `ruff check --fix` act as a reformatter, and so I hoped it was suitable to be included in Darker.  
But if you say Graylint will work well with it, I don't mind running the extra command 😉   
I guess I'll just have to try it out and see...

src/darker/formatters/ruff_formatter.py
akaihola: I reworked the whole `_get_supported_target_versions()` function, hope you like it better now.

src/darker/formatters/ruff_formatter.py
akaihola:
```python
>>> your_guy = "Arthur Dent"
>>> print(f"Your guy is {your_guy}")
Your guy is Arthur Dent
>>> print(f"Your guy is {your_guy!r}")
Your guy is 'Arthur Dent'
```

From Python's [f-strings] documentation:
>If a conversion is specified, the result of evaluating the expression is converted before formatting. Conversion `'!s'` calls [`str()`](https://docs.python.org/3/library/stdtypes.html#str "str") on the result, `'!r'` calls [`repr()`](https://docs.python.org/3/library/functions.html#repr "repr"), and `'!a'` calls [`ascii()`](https://docs.python.org/3/library/functions.html#ascii "ascii").

[f-strings]: https://docs.python.org/3/reference/lexical_analysis.html#formatted-string-literals

clintonsteiner: Ok that is pretty handy, thank you for the explanation on this one

akaihola:
>However, I was referring specifically to the --fix option of ruff check, which not only reports linting errors but also fixes them - which means modifying the code.

@zmeir, thanks for clarifying and sorry for not catching your intention right away.

Yes, modifying the code falls into Darker's realm of course. If Ruff does code modifications which reorder, move, remove or insert code, the implementation and reliability will be a bit tricky. For simple small in-place modifications probably another simple formatter backend will do just fine – I'm right now working on one for [pyupgrade].

Could you draft a separate GitHub issue for `ruff check --fix` support?

[pyupgrade]: /asottile/pyupgrade

zmeir:
> Could you draft a separate GitHub issue for `ruff check --fix` support?

Sure! Please see #758

src/darker/__main__.py
akaihola: We now have
```python
    logger.debug(
        "Running %r by %s.%s resulted in %s changed lines within a total of %s lines",
        formatter.name,
        formatter.__module__,
        type(formatter).__name__,
        "no" if formatted == fstringified else "some",
        len(formatted.lines),
    )
```

src/darker/files.py
```diff
@@ -25,22 +16,116 @@
...
+    r"|\.hg"
+    r"|\.ipynb_checkpoints"
+    r"|\.mypy_cache"
+    r"|\.nox"
+    r"|\.pytest_cache"
+    r"|\.ruff_cache"
+    r"|\.tox"
+    r"|\.svn"
+    r"|\.venv"
+    r"|\.vscode"
```
jedie, line 40: Maybe exclude all dot paths: `.*` ?

src/darker/formatters/black_formatter.py
```diff
@@ -96,6 +96,27 @@ def read_config(self, src: tuple[str, ...], args: Namespace) -> None:
...
+                " `pip install 'darker[black]'` or"
+                " `pip install black`"
+            )
+            logger.warning(
+                "To use a different formatter or no formatter, select it on the"
+                " command line (e.g. `--formatter=none`) or configuration"
+                " (e.g. `formatter=none`)"
+            )
+            message = "Can't find the Black package"
+            raise DependencyError(message) from exc
```
jedie, line 118: What will happen, if the internals of black will change? e.g.: renaming/moving of `parse_pyproject_toml` or `re_compile_maybe_verbose` ? Think the error message is not really good in this case.

So, what's about to import only black package and then do the real imports? 

e.g.:
```
        try:
            import black
        except ImportError as exc:
            logger.warning(
                "To re-format code using Black, install it using e.g."
                " `pip install 'darker[black]'` or"
                " `pip install black`"
            )
            logger.warning(
                "To use a different formatter or no formatter, select it on the"
                " command line (e.g. `--formatter=none`) or configuration"
                " (e.g. `formatter=none`)"
            )
            message = "Can't find the Black package"
            raise DependencyError(message) from exc

        # Black is installed
        from black import (  # pylint: disable=import-outside-toplevel
            parse_pyproject_toml,
            re_compile_maybe_verbose,
        )
```

src/darker/formatters/black_formatter.py
```diff
@@ -96,6 +96,27 @@ def read_config(self, src: tuple[str, ...], args: Namespace) -> None:
...
+        # Local import so Darker can be run without Black installed.
+        # Do error handling here. This is the first Black importing method being hit.
+        try:
+            from black import (  # pylint: disable=import-outside-toplevel
+                parse_pyproject_toml,
+                re_compile_maybe_verbose,
+            )
+        except ImportError as exc:
+            logger.warning(
+                "To re-format code using Black, install it using e.g."
```
jedie, line 108: Will this warning will be always displayed to the user?

src/darker/files.py
akaihola: My thought was to follow what Black does as closely as possible unless there's a good reason.

[black/src/black/const.py]:
```python
DEFAULT_EXCLUDES = r"/(\.direnv|\.eggs|\.git|\.hg|\.ipynb_checkpoints|\.mypy_cache|\.nox|\.pytest_cache|\.ruff_cache|\.tox|\.svn|\.venv|\.vscode|__pypackages__|_build|buck-out|build|dist|venv)/"  # noqa: B950
```

[black/src/black/const.py]: https://github.com/psf/black/blob/fdabd424e237f89d8b3f53cf8157902bbf850e45/src/black/const.py#L2

src/darker/formatters/black_formatter.py
akaihola: Thanks @jedie, I put this in a separate PR #792

src/darker/formatters/black_formatter.py
```diff
@@ -173,19 +196,23 @@ def _make_black_options(self) -> Mode:
         # Collect relevant Black configuration options from ``self.config`` in order to
         # pass them to Black's ``format_str()``. File exclusion options aren't needed
         # since at this point we already have a single file's content to work on.
+
+        # Local import so Darker can be run without Black installed.
+        # No need for error handling, already done in `BlackFormatter.read_config`.
+        from black import FileMode as Mode  # pylint: disable=import-outside-toplevel
+        from black import TargetVersion  # pylint: disable=import-outside-toplevel
```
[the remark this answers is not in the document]
akaihola: Thanks @jedie for the idea, a `black_wrapper.py` module has now been created, and I'm doing the same in #755 for the Pyupgrade plugin.

src/darker/formatters/black_formatter.py
akaihola: This question has been addressed in #792.

src/darker/files.py
akaihola: `gen_python_files` has now been copied into Darker's own code base. The reason for this was to remove the hard dependency on Black. See #759. So the code above isn't needed any more and has been removed.

src/darker/files.py
akaihola: Comments added in #798.

src/darker/files.py
akaihola: Implemented in #798. Note that instead of `{}`, `set()` needed to be used as the initial value.

PR #751, opened 2024-10-02
Respect isort's `skip_glob` setting

### LINKED ISSUE #615: `isort`: `skip_glob` option doesn't work
**Describe the bug**
We have Django files/settings that have intentional import ordering to avoid issues. Hence, we exclude them when fixing the imports using `isort` via `skip_glob` option. 

This is the relevant setting in `pyproject.toml`:
```toml
[tool.darker]
isort = true  # Run the `isort` also when running `darker`.

[tool.isort]
# https://pycqa.github.io/isort/docs/configuration/options.html
skip_glob = ['*/conf/settings/*']  # Exclude the Settings files due to sensitive imports.
filter_files = true  # Exclude the file if applicable even if it's explicitly specified. 
```

Using `isort` works:
```bash
$ isort conf/settings/backoffice.py --diff
Skipped 1 files
```

Using `darker` will still show the diff though which is unexpected:
```bash
$ darker conf/settings/backoffice.py --diff
...
```

**To Reproduce**
Basically, edit a Python file that is part of the `skip_glob`. Notice that `darken` will not exclude it.

**Expected behavior**
Just like the stand-alone `isort`, `darken --isort` should exclude the file since it's skipped via glob.

**Environment**
 - OS: macOS 12.5 Monterey
 - Python version: 3.11.0
 - Git version: 2.39.1
 - Darker version: 2.1.1
 - Black version: 24.4.2
 - Isort version: 5.13.2

Note that when using `darker` via `pre-commit`, it's the same issue: those supposed to be skipped files via glob pattern are still being linted by the `--isort` option.

akaihola:
Fixes #615.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,4 +7,5 @@ __pycache__/
 /.vscode/
 /build
 /venv
+/.venv
 uv.lock
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -20,6 +20,7 @@ Added
   Black support.
 - pyupgrade_ is now supported as a formatter plugin. Note that changes from pyupgrade
   are applied on a per-file basis, not only for modified lines as with Black_ and Ruff_.
+- Respect the ``skip_glob`` setting of isort_.
 
 Removed
 -------
@@ -28,6 +29,7 @@ Removed
 - In the Darker configuration file under ``[tool.darker]``, the Black configuration
   options ``skip_string_normalization`` and ``skip_magic_trailing_comma`` are no longer
   valid. Use ``[tool.black]`` instead.
+- Drop support for isort_ 5.1.x.
 
 Fixed
 -----
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,7 +39,7 @@ flynt = [
     "flynt>=0.76",
 ]
 isort = [
-    "isort>=5.1.0",
+    "isort>=5.2.0",
 ]
 ruff = [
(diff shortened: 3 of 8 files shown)


akaihola:
@ranelpadon I started by successfully adding a failing unit test. Next, looking deeper into the reason for the wrong behavior. It's somehow related to the fact that Darker calls `isort`'s API directly instead of running it as a subprocess.

I invited you as a collaborator on this repository, hoping you could be the reviewer for this PR. Would you have time for that?

akaihola:
Ok @ranelpadon there seem to be two reasons for `isort` ignoring `skip_glob =`:
1. a bug in passing the directory of the configuration file to `isort`
2. we pass Python code directly to `isort.code` instead of letting it read the file from the disk (see [darker.import_sorting:137](/akaihola/darker/blob/e854b7415e62ba15dd85500b6d0976e5ed71af87/src/darker/import_sorting.py#L137)), and in this case `isort` simply disregards the `skip_glob =` setting (see [isort.api:188](/PyCQA/isort/blob/7de182933fd50e04a7c47cc8be75a6547754b19c/isort/api.py#L188-L189))

Reason 1. is easily fixed, but 2. is a trickier one. I'll need to spend a little more time on it.

akaihola:
Not so difficult to fix after all. @ranelpadon could you verify that this branch works for you?

This PR is best reviewed one commit at a time.

Issue #78, opened 2020-09-12
Generate and publish documentation


akaihola:
Black has Sphinx documentation [auto-generated](https://github.com/psf/black/blob/20.8b1/docs/conf.py#L111) from the extensive [README.md](https://github.com/psf/black/blob/master/README.md) file. The documentation is then published on [readthedocs.io](black.readthedocs.io/).

We could do the same, but since our [README.rst](https://github.com/akaihola/darker/blob/master/README.rst) is ReStructuredText like Sphinx documentation by default, we could use the [`.. include::`](https://docutils.sourceforge.io/docs/ref/rst/directives.html#including-an-external-document-fragment) directive with the ``start-after`` and ``end-before`` options to include parts of the README in correct documentation sections, instead of running custom code to extract the text.

PR #764, opened 2024-10-22
create base for readthedocs sphinx site

clintonsteiner:
Generate with
```bash
sphinx-build -b html ./source ./build
```
or
```
make html
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,20 @@
+# Minimal makefile for Sphinx documentation
+#
+
+# You can set these variables from the command line, and also
+# from the environment for the first two.
+SPHINXOPTS    ?=
+SPHINXBUILD   ?= sphinx-build
+SOURCEDIR     = source
+BUILDDIR      = build
+
+# Put it first so that "make" without argument is like "make help".
+help:
+	@$(SPHINXBUILD) -M help "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
+
+.PHONY: help Makefile
+
+# Catch-all target: route all unknown targets to Sphinx using the new
+# "make mode" option.  $(O) is meant as a shortcut for $(SPHINXOPTS).
+%: Makefile
+	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
--- a/source/index.rst
+++ b/source/index.rst
@@ -1,0 +1,114 @@
+Welcome to Darker's Documentation!
+===================================
+
+|build-badge| |license-badge| |pypi-badge| |downloads-badge| |black-badge| |changelog-badge|
+
+.. |build-badge| image:: https://github.com/akaihola/darker/actions/workflows/python-package.yml/badge.svg
+   :alt: master branch build status
+   :target: https://github.com/akaihola/darker/actions/workflows/python-package.yml?query=branch%3Amaster
+.. |license-badge| image:: https://img.shields.io/badge/License-BSD%203--Clause-blue.svg
+   :alt: BSD 3 Clause license
+   :target: https://github.com/akaihola/darker/blob/master/LICENSE.rst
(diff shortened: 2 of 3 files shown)


clintonsteiner:
Linking [https://github.com/akaihola/darker/issues/78](url)

Makefile
```diff
@@ -0,0 +1,20 @@
+# Minimal makefile for Sphinx documentation
```
akaihola, line 1: Would it be more modern to use [just] and `justfile`?

[just]: https://just.systems/man/en/

akaihola: Ah, but Sphinx has `Makefile` support out of the box, and probably no support for `just`...

source/index.rst
```diff
@@ -0,0 +1,114 @@
...
+   :target: https://github.com/akaihola/darker/blob/master/LICENSE.rst
+.. |pypi-badge| image:: https://img.shields.io/pypi/v/darker
+   :alt: Latest release on PyPI
+   :target: https://pypi.org/project/darker/
+.. |downloads-badge| image:: https://pepy.tech/badge/darker
+   :alt: Number of downloads
+   :target: https://pepy.tech/project/darker
+.. |black-badge| image:: https://img.shields.io/badge/code%20style-black-000000.svg
+   :alt: Source code formatted using Black
+   :target: https://github.com/psf/black
```
akaihola, line 20: Why not de-duplicate these, too, using `:start-at:` and `:end-before:`?

source/index.rst
```diff
@@ -0,0 +1,114 @@
...
+.. |changelog-badge| image:: https://img.shields.io/badge/-change%20log-purple
+   :alt: Change log
+   :target: https://github.com/akaihola/darker/blob/master/CHANGES.rst
+
+What is Darker?
+===============
+
+.. include:: ../README.rst
+   :start-after: What?
+   :end-before: Why?
```
akaihola, line 30: Could we copy the titles as well, like:
```suggestion
.. include:: ../README.rst
   :start-at: What?
   :end-before: Why?
```

akaihola: Or is it that we don't get proper references and tables of content if the headings are not actually literally present in the Sphinx document?

akaihola (changes requested):
In `setup.cfg`, we should probably add `sphinx` to either `darker[release]` extra dependencies, or to a new `darker[docs]` extra.

Makefile
clintonsteiner: I'll be honest this is the first I've heard of just and justfile, I do like the support out of the box and know how this works with make if you're good with that

source/index.rst
clintonsteiner: image
Getting errors doing that way

source/index.rst
clintonsteiner: Right - we don't have anything to reference to doing that

clintonsteiner:
https://github.com/akaihola/darker/pull/766 
Think this is probably a better approach, thoughts?

akaihola:
> #766 Think this is probably a better approach, thoughts?

#766 is indeed way simpler and adds minimal changes. It's a better approach as long as we choose to stick with a single page of documentation.

I'm trying to think of any benefits from full-fledged multi-page documentation with a more focused README. The change log and contribution guide could then of course be rendered as part of the documentation pages, and maybe more technical detail included as a separate page.

What do you think about the trade-offs between the two approaches?

clintonsteiner:
> > #766 Think this is probably a better approach, thoughts?
> 
> 
> 
> #766 is indeed way simpler and adds minimal changes. It's a better approach as long as we choose to stick with a single page of documentation.
> 
> 
> 
> I'm trying to think of any benefits from full-fledged multi-page documentation with a more focused README. The change log and contribution guide could then of course be rendered as part of the documentation pages, and maybe more technical detail included as a separate page.
> 
> 
> 
> What do you think about the trade-offs between the two approaches?

I think choose the simple option until theres a reason to do the hard one

PR #766, opened 2024-10-31
README sphinx website generation

clintonsteiner:
* Made significantly easier and still used sphinx

---

Notes:
- This simpler approach than #764 works as long as we choose to stick with a single page of documentation. If we e.g. wanted to render the change log and contribution guide as part of the documentation pages, the approach drafted in #764 should be used instead.
  – @akaihola 2024-11-16


.github/actions/SphinxBuild.yml
```diff
@@ -0,0 +1,35 @@
...
+      - master
+
+jobs:
+  sphinx-build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.9'
```
akaihola, line 16: I'm curious, why build with an old Python version?

clintonsteiner: Fixed! Good callout

.github/actions/SphinxBuild.yml
akaihola, line 1: Doesn't this workflow file belong in `.github/workflows/` instead?

clintonsteiner: Fixed this

.github/workflows/sphinx-build.yml
```diff
@@ -0,0 +1,35 @@
...
+        run: ./action/build_docs.sh
+
+      - name: Upload artifacts
+        uses: actions/upload-artifact@v1
+        with:
+          name: html-docs
+          path: ./sphinx_output/
+
+      - name: Deploy
+        uses: peaceiris/actions-gh-pages@v3
```
akaihola, line 30: Just to be sure, so this is going to publish the documentation at https://akaihola.github.io/darker?

akaihola (commented):
Looking at https://clintonsteiner.github.io/darker-docs/ it seems limiting the TOC to top level only would be an improvement. Now the detailed TOC fills the top screenful of the documentation page.

PR #776, opened 2024-11-25
fix dependabot - can't live in workflows dir

clintonsteiner:


clintonsteiner:
@akaihola https://stackoverflow.com/questions/69446872/dependabot-error-githubl1-no-event-triggers-defined-in-on
This will fix current issues

akaihola (approved):

PR #802, opened 2025-01-06
Update verification.py per pyupgrade parsing

clintonsteiner:
Use built in dict type instead of typing Dict

### DIFF
--- a/src/darker/verification.py
+++ b/src/darker/verification.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import ast
 import sys
 import warnings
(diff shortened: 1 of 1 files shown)


clintonsteiner:
@akaihola please direct your attention here first as this will fix checks for several other reviews

akaihola (commented):
@clintonsteiner the Pyupgrade workflow currently indeed wants to update the whole code base. I'd like to hold on with its suggested changes until #755 is merged. After that, we can dogfeed Darker's own code base to the new Pyupgrade workflow which runs Pyupgrade through Darker and only updates modified files.

PR #803, opened 2025-01-08
Use packaging.version.Version in test-future.yml

akaihola:
Dropping use of `distutils` which wasn't explicitly required anyway, but came in through `pip`.

### DIFF
--- a/.github/workflows/test-future.yml
+++ b/.github/workflows/test-future.yml
@@ -40,9 +40,9 @@ jobs:
           import json
           import os
           import urllib.request
-          from distutils.version import LooseVersion
-          from importlib.metadata import version
           from textwrap import dedent
+
+          from packaging.version import Version
 
           for linenum, line in enumerate(open("setup.cfg"), 1):
               constraint = line.strip()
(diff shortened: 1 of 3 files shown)


.github/workflows/test-future.yml
```diff
@@ -40,10 +40,10 @@ jobs:
           import json
           import os
           import urllib.request
-          from distutils.version import LooseVersion
-          from importlib.metadata import version
           from textwrap import dedent
```
clintonsteiner, line 44:
```suggestion
```

clintonsteiner: nit: extra blank line

.github/workflows/test-future.yml
```diff
@@ -40,10 +40,10 @@ jobs:
           import json
           import os
           import urllib.request
```
clintonsteiner, line 42:
```suggestion
```

clintonsteiner: Believe this is unused as well

clintonsteiner (approved):

.github/workflows/test-future.yml
akaihola: It's a blank line for grouping imports into standard library ones and 3rd party ones.

.github/workflows/test-future.yml
akaihola: There's this on line 56:
```python
          response = urllib.request.urlopen(
              'https://pypi.org/pypi/black/json'
          ).read().decode()
```

PR #810, opened 2025-01-18
uv work: implement oldest dependencies into pyproject.toml

### LINKED ISSUE #806: Move from Setuptools to a pyproject.toml based build tool
Moving away from Setuptools hasn't yet been possible due to [dynamic removal of the contributors HTML table](https://github.com/akaihola/darker/blob/c0571f3c4e1ebfa08d8b5422d55bf9425636c9f2/setup.py#L40-L48) from `README.rst` in `setup.py`. It's necessary because PyPI can't render it correctly. Let's find a solution using more modern `pyproject.toml` based build tools.

_Originally posted by @akaihola in https://github.com/akaihola/darker/issues/773#issuecomment-2576969030_

clintonsteiner:
Fixes #806


.github/workflows/python-package.yml
```diff
@@ -24,7 +24,9 @@ jobs:
         uses: astral-sh/setup-uv@v3
       - uses: actions/setup-python@v5
       - name: Build wheel distribution
-        run: uv build --wheel
+        run: |
+          uv sync --all-packages
```
akaihola, line 28: Why is `uv sync --all-packages` needed before running `uv build`?

Seems unrelated to testing with oldest dependencies?

.github/workflows/python-package.yml
```diff
@@ -24,7 +24,9 @@ jobs:
...
+          uv build
```
akaihola, line 29: Will `uv build` without `--heel` also unnecessarily build the source distribution?

This change seems unrelated to testing with oldest dependencies.

.github/workflows/python-package.yml
```diff
@@ -59,7 +61,7 @@ jobs:
         with:
           nix_path: nixpkgs=channel:nixos-unstable
       - name: Download wheel uploaded by the build-wheel job
-        uses: actions/download-artifact@v4.1.7
+        uses: actions/download-artifact@v4
```
akaihola, line 64: I extracted `{upload,download}-artifact@v4` changes into #812. Let's merge that one, then rebase.

.github/workflows/python-package.yml
```diff
@@ -105,38 +107,35 @@ jobs:
...
       - name: Upgrade Black, Flynt and Isort from GitHub if future constraints are used
         # This can't be done in the same Pip invocation as installing Darker
         # since Darker might place an upper limit on Black, Flynt and/or Isort during
         # compatibility fixing periods.
         if: matrix.post_install
         run: ${{ matrix.post_install }}
-      - name: Downgrade target-version when running oldest supported Black
-        if: matrix.constraints == '--constraint constraints-oldest.txt'
-        run: |
-          sed -i 's/, "py311", "py312"//' pyproject.toml
```
akaihola, line 139: This was required since we used to also run the "oldest" tests using the oldest supported Python, and needed to instruct Black to format according to rules compatible with that version.

Is this no longer necessary?

.github/workflows/test-future.yml
```diff
@@ -17,22 +17,14 @@ jobs:
         with:
           ref: 'master'
           fetch-depth: 0
-      - uses: actions/setup-python@v5
```
akaihola, line 20: Letting `uv` install Python?

[uv documentation](https://docs.astral.sh/uv/guides/integration/github/#setting-up-python) says:
>Alternatively, the official GitHub `setup-python` action can be used. This can be faster, because GitHub caches the Python versions alongside the runner.

uv.lock
```diff
@@ -0,0 +1,1382 @@
+version = 1
```
akaihola, line 1: To integrate well with different development workflows, I think we should do our best effort to enable installing Darker and its dependencies in project environments. For that reason, it's good to support as many versions of dependencies as possible.

To enable easy testing against different versions of dependencies, I would advocate for leaving `uv.lock` out from the repository so we don't enforce any versions in development environments.

See [The lockfile](https://docs.astral.sh/uv/concepts/projects/layout/#the-lockfile) in uv documentation.

akaihola (changes requested):
I ran `uv build` in this branch, and now `README.rst` in the source distribution contains the HTML table for contributors. This is not desired since PyPI is not able to render raw HTML (see build [#2766](https://github.com/akaihola/darker/actions/runs/12847633639/job/35824141186?pr=810)):
```shell
Checking dist/darker-2.1.1-py3-none-any.whl: FAILED
ERROR    `long_description` has syntax errors in markup and would not be        
         rendered on PyPI.                                                      
         line 974: Warning: "raw" directive disabled.                           
Checking dist/darker-2.1.1.tar.gz: FAILED
ERROR    `long_description` has syntax errors in markup and would not be        
         rendered on PyPI.                                                      
         line 974: Warning: "raw" directive disabled. 
```

Before migrating from `setup.cfg` to `pyproject.toml`, we need to solve this somehow. We want the contributor table to appear on GitHub but not on PyPI – unless the rendering issue can be solved.

The dynamic README in `pyproject.toml` just reads `README.rst` as such and doesn't run the code in `setup.py` for removing the contributor table. See [Dynamic Metadata](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata) documentation for Setuptools.

This PR essentially fixes #806 which also explains the README issue.

.github/workflows/python-package.yml
```diff
@@ -105,38 +107,35 @@ jobs:
...
           - os: ubuntu-latest
             python-version: '3.9'
-            constraints: '--constraint constraints-oldest.txt'
+            post_install: uv sync --resolution lowest-direct 
           - os: ubuntu-latest
             python-version: '3.12'
-            post_install: pip uninstall -y -r constraints-future.txt ;
-                          pip install -r constraints-future.txt
-                                      --upgrade --upgrade-strategy=eager
+            post_install: uv sync --upgrade --latest
```
akaihola, line 113: `--latest` doesn't seem to be a valid `uv` option, see build [#2766](https://github.com/akaihola/darker/actions/runs/12847633639/job/35824142557?pr=810):
```
Run uv sync --upgrade --latest
  uv sync --upgrade --latest
  shell: /usr/bin/bash -e {0}
  env:
    PY_IGNORE_IMPORTMISMATCH: 1
    UV_CACHE_DIR: /home/runner/work/_temp/setup-uv-cache
    pythonLocation: /opt/hostedtoolcache/Python/3.12.8/x64
    PKG_CONFIG_PATH: /opt/hostedtoolcache/Python/3.12.8/x64/lib/pkgconfig
    Python_ROOT_DIR: /opt/hostedtoolcache/Python/3.12.8/x64
    Python2_ROOT_DIR: /opt/hostedtoolcache/Python/3.12.8/x64
    Python3_ROOT_DIR: /opt/hostedtoolcache/Python/3.12.8/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.12.8/x64/lib
error: unexpected argument '--latest' found

Usage: uv sync --upgrade

For more information, try '--help'.
Error: Process completed with exit code 2.
```

pyproject.toml
```diff
@@ -2,6 +2,99 @@
 # Minimum requirements for the build system to execute.
 requires = ["setuptools", "wheel"]  # PEP 508 specifications.
 
+[tool.setuptools.dynamic]
+readme = {file=["README.rst"]}
+
```
akaihola, line 7: This section needs to be removed. Otherwise README modifications done by `setup.py` won't be used. See [this comment] from #806.

```suggestion
```

[this comment]: /akaihola/darker/issues/806#issuecomment-2629563755

pyproject.toml
```diff
@@ -2,6 +2,99 @@
...
+[project]
```
akaihola, line 8: This section needs to be removed. Otherwise README modifications done by `setup.py` won't be used. See [this comment] from #806.

```suggestion
[project]
```

[this comment]: /akaihola/darker/issues/806#issuecomment-2629563755

.github/workflows/python-package.yml
clintonsteiner: agreed, changed back

.github/workflows/python-package.yml
clintonsteiner: Rebased

.github/workflows/python-package.yml
clintonsteiner: That will still be necessary - putting back

uv.lock
clintonsteiner: removed

.github/workflows/python-package.yml
clintonsteiner: Appears the resolution strategy is latest by default - so will remove

akaihola (dismissed):
One more conversation to resolve regarding the call of `uv sync --all-packages` in `python-package.yml`.

.github/workflows/python-package.yml
clintonsteiner: Removed

akaihola (approved):

PR #846, opened 2025-09-08
Fix missing 'v' in github action example

k-dominik:
will not find the version otherwise...


akaihola:
Thanks @k-dominik for the fix! Could you add a note under the `Fixed` section in `README.rst` as well?

k-dominik:
done!

PR #848, opened 2025-09-17
Use Python to invoke ruff instead of binary

anakinxc:
When user installs `ruff` via `pip install --user`, `ruff` binary is usually not on system path.

Directly invoke `ruff` will cause error `No such file or directory: 'ruff'`, ask Python to invoke the python one is an safer approach.

### DIFF
--- a/src/darker/formatters/ruff_formatter.py
+++ b/src/darker/formatters/ruff_formatter.py
@@ -228,6 +228,8 @@ def _ruff_format_stdin(
 
     """
     cmdline = [
+        sys.executable,
+        "-m",
         "ruff",
         "format",
         "--force-exclude",  # apply `exclude =` from conffile even with stdin


tpppppub:
Hit the same issue

akaihola:
Thanks @anakinxc and @tpppppub!

Won't this then break `--formatter=ruff` for those who like to install Ruff using an operating system package manager or manually?

If that's the case, what options do we have? The following come to mind:
- first see if the `ruff` Python package is installed, and only use `python -m ruff` if that's the case, otherwise just `ruff`
- first see if the `ruff` binary is found in `PATH`, and use just `ruff` if that's the case, otherwise try `python -m ruff`
- just use `ruff`, and if it fails, try `python -m ruff`
- use `python -m ruff`, and if it fails, try just `ruff`
- ...other?

anakinxc:
> * just use `ruff`, and if it fails, try `python -m ruff`

Personally, I think this one is a better solution

